PR #1, opened 2018-10-17
Initial commit

JohnCleese:
Dockerfile has been added
docker-compose.yml has been added
requirements has been added
wait-for-it has been added
requirements.txt has been added


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
magul: `.idea` directory is PyCharm specific and should not be stored in git repository. Use `.gitignore` to ingore it.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ADD ./wait-for-it.sh /usr/bin/wait-for-it
+RUN chmod a+x /usr/bin/wait-for-it
+
+# Backend setup
+RUN pip3 install --upgrade pip
+
+# Add build-time directories
+ADD . /pah-fm
+WORKDIR /pah-fm
+RUN pip3 install -r ./requirements.txt
```
magul: remember to always add newline at the end of a file

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "2"
```
magul: As we spoke before, we should investigate and resolve an issue, why do we need to use `2` instead of `3` for Docker syntax version.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+services:
+
+  backend:
+    build: .
+    command: tail -f /dev/null
+    environment:
+      DJANGO_SETTINGS_MODULE: pah-fm_org.settings.dev
```
magul: Do we use these variables?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      PAH-FM_SECRET_KEY: a63eb5ef-3b25-4595-846a-5d97d99486f0
+      PAH-FM_DB_HOST: db
+      PAH-FM_DB_PASS: pah-fm
+    ports:
+      - "8000:8000"
+    links:
+      - db:db
+    volumes:
+      - ./backend:/pah-fm/backend
+      - /volontulo/backend/apps/pah-fm/node_modules
```
magul: `/volontulo` ???

requirements.txt
```diff
@@ -0,0 +1,12 @@
+Django==2.1.2
```
magul: Do we need all these packages? Can You pick only these, that we actually going to use?

docker-compose.yml
magul: As You haven't initialize Django project none of these variables is usable here.

requirements.txt
magul: There's no Django initialize, so maybe leave here only `Django`?

wait-for-it.sh
```diff
@@ -0,0 +1,177 @@
...
+
+if [[ $CLI != "" ]]; then
+    if [[ $RESULT -ne 0 && $STRICT -eq 1 ]]; then
+        echoerr "$cmdname: strict mode, refusing to execute subprocess"
+        exit $RESULT
+    fi
+    exec "${CLI[@]}"
+else
+    exit $RESULT
+fi
```
magul: newline at the end of a file

PR #22, opened 2018-10-28
Django models cleanup

magul:


backend/fleet_management/models.py
```diff
@@ -3,31 +3,31 @@
...
     def __str__(self):
         return self.plates
 
 
-class PassengerModel(models.Model):
-    first_name = models.CharField(max_length=20, blank=False, unique=True)
-    last_name = models.CharField(max_length=20, blank=False, unique=True)
+class Passenger(models.Model):
+    first_name = models.CharField(max_length=60, blank=False, unique=True)
+    last_name = models.CharField(max_length=60, blank=False, unique=True)
```
juliaanholcer: Is `unique=True` a good idea?

magul: Haven’t seen that. Good catch!

backend/fleet_management/models.py
```diff
@@ -3,31 +3,31 @@
...
         return self.last_name
 
 
-class RouteModel(models.Model):
-    driver = models.ForeignKey(DriverUserProfile, on_delete=models.CASCADE)
-    car = models.ForeignKey(CarModel, null=False, on_delete=models.CASCADE)
-    passengers = models.ManyToManyField(PassengerModel)
+class Route(models.Model):
+    driver = models.ForeignKey(Driver, on_delete=models.CASCADE)
+    car = models.ForeignKey(Car, null=False, on_delete=models.CASCADE)
```
jacol12345: More of a product requirement: do we want to delete drivers and cars? I think "disabling" them would be better - we may need the data later.

magul: Most probably we will disable them, but Django since version 2.x requires `on_delete` parameter.

magul:
@juliaanholcer your comment was fixed, @jacol12345 your one was answered to.

PR #26, opened 2018-11-03
Work ksawery

JohnCleese:
Fuel consumption added

Cars and passengers list endpoints added (no auth yet)


backend/fleet_management/models.py
```diff
@@ -33,3 +33,8 @@ class Route(models.Model):
     end_mileage = models.IntegerField(null=False)
     fuel_level_begin = models.FloatField(null=False)
     fuel_level_end = models.FloatField(null=False)
+
+    def fuel_consumption(self):
```
jacol12345: Please make it either a property by decorating with `@property` OR change method name to `get_fuel_consumption`. It will be easier to reason whether it's a prop or method by name this way.

backend/fleet_management/models.py
```diff
@@ -33,3 +33,8 @@ class Route(models.Model):
...
+        fuel_delta = self.fuel_level_end - self.fuel_level_begin
+        mileage_delta = self.end_mileage - self.start_mileage
+        return 100*(fuel_delta/mileage_delta)
```
jacol12345: Please add spaces around operators
https://www.python.org/dev/peps/pep-0008/#other-recommendations

backend/pah_fm/urls.py
```diff
@@ -15,7 +15,10 @@
 """
 from django.contrib import admin
 from django.urls import path
+from ..fleet_management.api import PassengerListView, CarListView
```
jacol12345: Does it work in Dockerized env? I had some trouble using relative import with top-level packages.

backend/pah_fm/urls.py
```diff
@@ -15,7 +15,10 @@
...
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('api/users/passengers', PassengerListView.as_view(), name='passengers'),
+    path('api/users/cars', CarListView.as_view(), name='cars'),
```
jacol12345: Why not `api/passengers` and `api/cars`? I think that passengers and cars are entities on their own.

backend/requirements.txt
```diff
@@ -0,0 +1,5 @@
+Django==2.1.2
```
jacol12345: We already have requirements directory with base and dev requirements. This file is redundant.

docker-compose.yml
```diff
@@ -16,6 +17,7 @@ services:
     # build: backend
     command: bash docker-entrypoint.sh
     # command: tail -f /dev/null
+
```
jacol12345: Can you elaborate why this empty line is necessary :)?

backend/requirements.txt
JohnCleese: My bad...

docker-compose.yml
JohnCleese: I have no idea :(

backend/fleet_management/models.py
JohnCleese: I didn`t know that. Thanks.

backend/fleet_management/models.py
JohnCleese: Yes, you are right.

magul:
Closing, as we discussed it witt @JohnCleese - he will create separate branches for every issue.

PR #42, opened 2018-11-08
PassengersList endpoint

JohnCleese:


backend/fleet_management/serializers.py
```diff
@@ -1,9 +1,18 @@
...
         model = User
         fields = ('id', 'username')
+
+
+class PassengerSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Passenger
+        fields = (
+            'first_name',
+            'last_name',
```
jacol12345: We need also passenger id to assign users to routes.

PR #43, opened 2018-11-08
CarList endpoint

JohnCleese:


backend/fleet_management/api.py
```diff
@@ -1,11 +1,26 @@
...
 
 class CurrentUserRetrieveView(views.APIView):
     def get(self, request):
         return Response(
             UserSerializer(request.user).data
         )
+
+
+class CarListView(generics.ListAPIView):
+    authentication_classes = (IsAuthenticated,)
```
jacol12345: I think this is default in current setup and therefore not required here.

backend/fleet_management/api.py
```diff
@@ -1,11 +1,26 @@
...
+    serializer_class = CarSerializer
+    queryset = Car.objects.all()
+    filter_backends = (filters.OrderingFilter, filters.SearchFilter, DjangoFilterBackend,)
```
jacol12345: Do we need both `SearchFilter` and `DjangoFilterBackend`?

backend/fleet_management/api.py
```diff
@@ -1,11 +1,26 @@
...
+    search_fields = (
```
jacol12345: I think that if we would have to search cars then the only field we will search by would be plates. However, it does not break anything so let's not close any 🚪 :)

backend/fleet_management/models.py
```diff
@@ -11,6 +11,9 @@ def __str__(self):
 
 class Car(models.Model):
     plates = models.CharField(max_length=10, blank=False, unique=True)
+    description = models.CharField(max_length=500, blank=True)
+    mileage_unit = models.CharField(max_length=10, blank=False)
```
jacol12345: This field needs choices https://docs.djangoproject.com/en/2.1/ref/models/fields/#choices

backend/fleet_management/serializers.py
```diff
@@ -1,9 +1,18 @@
...
         model = User
         fields = ('id', 'username')
+
+
+class CarSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Car
+        fields = (
+            'plates',
+            'fuel_consumption',
```
jacol12345: Let's return all fields the car has, including id.

jacol12345:
I'm closing the ticket as it has just be done.

PR #44, opened 2018-11-08
Projects list endpoint

JohnCleese:


backend/fleet_management/api.py
```diff
@@ -1,11 +1,25 @@
-from rest_framework import views
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import generics, filters, views
 from rest_framework.response import Response
-
-from .serializers import UserSerializer
+from .serializers import UserSerializer, ProjectSerializer
+from .models import Project
+from rest_framework.permissions import IsAuthenticated
```
jacol12345: Please keep imports sorted and grouped by:
* Python builtins
* 3rd party libraries (like django, rest_framework)
* our modules

backend/fleet_management/api.py
```diff
@@ -1,11 +1,25 @@
...
         return Response(
             UserSerializer(request.user).data
         )
+
+
+class ProjectsListView(generics.ListAPIView):
+    authentication_classes = (IsAuthenticated,)
+    serializer_class = ProjectSerializer
+    queryset = Project.objects.all()
+    filter_backends = (filters.OrderingFilter, filters.SearchFilter, DjangoFilterBackend,)
```
jacol12345: Do we need both `filters.OrderingFilter` and `DjangoFilterBackend`?

backend/fleet_management/models.py
```diff
@@ -33,3 +33,12 @@ class Route(models.Model):
     end_mileage = models.IntegerField(null=False)
     fuel_level_begin = models.FloatField(null=False)
     fuel_level_end = models.FloatField(null=False)
+
+
+class Project(models.Model):
+    routes = models.ManyToManyField(Route)
+    project_title = models.CharField(max_length=50, blank=False)
```
jacol12345: This field could be named just `title`, there is no need to repeat `project`.

backend/fleet_management/api.py
```diff
@@ -1,11 +1,25 @@
...
+    search_fields = (
+        'project_title',
```
jacol12345: Let's make it searchable by title only to keep things simple.

backend/fleet_management/serializers.py
```diff
@@ -1,9 +1,19 @@
...
         fields = ('id', 'username')
+
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = (
+            'project_title',
+            'project_description',
+            'routes',
```
jacol12345: We don't need routes returned from Project.

backend/fleet_management/api.py
```diff
@@ -1,11 +1,22 @@
...
 
 class CurrentUserRetrieveView(views.APIView):
     def get(self, request):
         return Response(
             UserSerializer(request.user).data
         )
+
+
+class ProjectsListView(generics.ListAPIView):
+    authentication_classes = (IsAuthenticated,)
```
jacol12345: It should be `permission_classes = (IsAuthenticated,)`

backend/fleet_management/api.py
```diff
@@ -1,11 +1,22 @@
...
+    serializer_class = ProjectSerializer
+    queryset = Project.objects.all()
+    filter_backends = (filters.OrderingFilter, filters.SearchFilter,)
+    search_fields = (
```
jacol12345: We will not use search for projects so we can remove `search_fields`.

backend/fleet_management/api.py
```diff
@@ -1,11 +1,22 @@
...
         return Response(
             UserSerializer(request.user).data
         )
+
+
+class ProjectsListView(generics.ListAPIView):
+    authentication_classes = (IsAuthenticated,)
+    serializer_class = ProjectSerializer
+    queryset = Project.objects.all()
+    filter_backends = (filters.OrderingFilter, filters.SearchFilter,)
```
jacol12345: We will not use search here so we can remove `SearchFilter` from filter_backends.

backend/fleet_management/api.py
```diff
@@ -1,11 +1,22 @@
...
     def get(self, request):
         return Response(
             UserSerializer(request.user).data
         )
+
+
+class ProjectsListView(generics.ListAPIView):
+    authentication_classes = (IsAuthenticated,)
+    serializer_class = ProjectSerializer
+    queryset = Project.objects.all()
```
jacol12345: Please add `ordering_fields` property  - it would be cool to get projects ordered by title ;)

jacol12345:
@JohnCleese I have given it some thought and it turns out that we don't need an endpoint with list of projects. I opened a PR with your work on Project model here: https://github.com/CodeForPoznan/pah-fm/pull/73.

Issue #35, opened 2018-11-07
List of routes view


jacol12345:
Add a link on main page to open the list of routes. This opens a list of routes of current user. They can browse it and expand for details.

Decide on the layout, make it simple but responsive and bootstraply good-looking.

PR #52, opened 2018-11-14
#35 List of routes view

juliaanholcer:
Add some routes via form to see routes on this view.

### DIFF
--- a/frontend/src/views/RouteFormView.vue
+++ b/frontend/src/views/RouteFormView.vue
@@ -89,6 +89,7 @@
 
 <script>
 import { mapActions } from 'vuex';
+import uuidv4 from 'uuid/v4';
 import * as actions from '../store/actions';
 
 export default {
@@ -96,6 +97,7 @@ export default {
   data() {
     return {
       route: {
+        id: '',
         date: '',
         description: '',
         from: '',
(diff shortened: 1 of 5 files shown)


jacol12345:
@juliaanholcer You can rebase the PR branch ;)

juliaanholcer:
@jacol12345 working on it :-) There's nothing better than some conflicts early in the morning.

juliaanholcer:
I wanted to make all of this on the computer with windows, and it was a really bad idea...

frontend/src/views/RouteFormView.vue
```diff
@@ -115,18 +117,21 @@ export default {
...
 
     validateForm() {
       this.errors = [];
 
-      Object.keys(this.route).forEach((key) => {
-        if (!this.route[key]) {
-          this.errors.push(`${key} is required`);
-        }
-      });
+      Object.keys(this.route)
```
jacol12345: If you want to make it super-nice :D 
```
this.errors = Object.entries(this.route).filter([key, val] => !val).map([key, val] => `${key} is required`);
```

jacol12345: No need to change, I was just boasting :)

PR #55, opened 2018-11-17
Add basic local storage route saving

adamgajzlerowicz:
This adds storing routes in local storage and rehydration of those upon app open.


frontend/src/views/RouteFormView.vue
```diff
@@ -92,27 +92,35 @@
...
+
+const defaultFormState = {
+  id: uuidv4(),
+  date: '',
+  description: '',
+  from: '',
+  destination: '',
+  startMileage: '',
+  endMileage: '',
+  isSynced: '',
```
jacol12345: default should be `false`

PR #56, opened 2018-11-18
Stylelint

adamgajzlerowicz:
This adds 
* stylelint rules for unified css style
* stylelint to pipeline

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,8 +6,9 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "http:server": "http-server dist -c-1",
-    "lint": "vue-cli-service lint --no-fix",
-    "lint:fix": "vue-cli-service lint"
+    "stylelint": "stylelint src/**/*.vue src/**/*.scss --syntax scss --cache",
+    "lint:fix": "vue-cli-service lint",
+    "lint": "vue-cli-service lint --no-fix && npm run stylelint"
   },
   "dependencies": {
     "@7ninjas/scss-mixins": "^1.0.1",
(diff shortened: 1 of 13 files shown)


frontend/package.json
```diff
@@ -6,19 +6,20 @@
...
     "bootstrap": "4.1.3",
     "bootstrap-vue": "2.0.0-rc.11",
     "jwt-decode": "^2.2.0",
     "register-service-worker": "^1.5.2",
+    "uuid": "3.3.2",
     "vue": "2.5.17",
     "vue-router": "3.0.1",
-    "vuex": "^3.0.1",
-    "uuid": "3.3.2"
+    "vuex": "^3.0.1"
```
weronikafelczak: We've discussed with @magul we shouldn't use "^", so maybe we could remove it here? :)

frontend/package.json
```diff
@@ -29,6 +30,9 @@
     "http-server": "^0.11.1",
     "node-sass": "4.9.0",
     "sass-loader": "7.0.1",
+    "stylelint": "^9.8.0",
+    "stylelint-config-standard": "^18.2.0",
+    "stylelint-processor-html": "^1.0.0",
```
weronikafelczak: And same here?

frontend/package.json
adamgajzlerowicz: `package-lock.json` takes care of that, although if this is a requirement, I am happy to comply. This PR doesn't change the version. NPM sorts imports alphabetically by default, so seems that it happened here too. 
If we are required to to use specific versions, lets make a separate pr that facilitates just that.

PR #59, opened 2018-11-20
added local storage sync

adamgajzlerowicz:
Includes
* sync of USER, ROUTES to local storage in a extensible way using vue-persisted state
* minor css tweaks: push footer to the bottom of the page, corrects nav to loop li elements instead of ul element


frontend/src/App.vue
```diff
@@ -48,4 +48,8 @@ export default {
   -moz-osx-font-smoothing: grayscale;
   color: #2c3e50;
 }
+
+.page-container {
+  min-height: calc(100vh - 65px - 128px);
```
jacol12345: How is that calculated, CSS Ninja :D?

frontend/src/components/Footer.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="footer p-3">
-    <h3><a href="http://codeforpoznan.pl">Code for Poznan</a></h3>
+    <h3 class="h3"><a href="http://codeforpoznan.pl">Code for Poznan</a></h3>
```
jacol12345: That seems a bit too much, why not style the h3 element?

jacol12345:
I need to run it before accepting - will do so tomorrow.

frontend/src/App.vue
adamgajzlerowicz: I take current height if header and current height of footer and subtract them from height of the screen. Currently those heights don't change on desktop, but I am waiting for the app to get connected to the api to see how many elements we will have on nav.
Once we are designing mobile we will likely update this too.

frontend/src/components/Footer.vue
adamgajzlerowicz: Using class names is always better idea. Gives more flexibility. Perhaps the name its-self is not that fortunate. I will change it to heading

PR #72, opened 2018-11-22
Fix Docker development tools.

jacol12345:
1) Add Nginx webserver service.
2) Use Nginx to proxy to Django.
3) Use Nginx to serve built assets.
4) Frontend container builds assets and puts them in shared
volume.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,10 +37,10 @@ services:
   webserver:
     image: nginx
     volumes:
-     - ./pah-fm.conf:/etc/nginx/nginx.conf
-     - asset-volume:/var/www/:ro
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -34,5 +33,16 @@ services:
...
       POSTGRES_PASSWORD: pah-fm
+
+  webserver:
+    image: nginx
+    volumes:
+     - ./pah-fm.conf:/etc/nginx/nginx.conf
+     - asset-volume:/var/www/:ro
     ports:
-      - "5432:5432"
+     - 8080:80
```
juliaanholcer: Is one space missing in indent here and in lines 40-41?

jacol12345: Good spot!

PR #80, opened 2018-11-28
Language switcher added with correct style.

lukaszmaterla:
Switcher working with EN, PL, UKR


frontend/src/components/Language.vue
```diff
@@ -0,0 +1,52 @@
...
+  name: 'Language',
+  components: {
+    En,
+    Pl,
+    Ua,
+  },
+  methods: {
+    changeLang(event) {
+      this._i18n.locale = languages[event];
+      console.log(this._i18n.locale);
```
jacol12345: Please remove console.log

frontend/src/translations.json
```diff
@@ -1,6 +1,77 @@
...
+    "common": {
+      "about": "Про",
+      "home": "Домашня сторінка",
+      "login": "Логін",
+      "new_route": "Додати новий маршрут",
+      "routes": "Маршрути",
+      "intro": "Польська гуманітарна акція - це польська неурядова організація, яка працює у Польщі та інших країнах. Його завдання \"зробити світ кращим місцем шляхом полегшення людських страждань та просування гуманітарних цінностей\"."
+    },
+    "routes": {
+      "please_correct_errors": "Please correct the following error(s):",
```
jacol12345: English is the default language. Other languages should only have the translated fields specified here, there is no need to repeat the English messages.

adamgajzlerowicz: I think that there is an error, where if keys are missing from translation, entire translation engine stops translating. It could be matter of configuration... not sure..

lukaszmaterla: Thanks, but anyway I gonna correct this in the next task

PR #84, opened 2018-11-29
#14 Drive: create endpoint

### LINKED ISSUE #14: Drives: create endpoint
Create an endpoint `/route` that only accept `POST` method and is responsible for creation of new Route (see `Route` model).

Accepts:
check out the model - it should extract driver from the request.

Available only to logged-in drivers.

jacol12345:
Closes #14.
Only happy path, minimal validation. 500 thrown when car does not exist and passengers will not be saved if wrong ids are supplied.


backend/fleet_management/serializers.py
```diff
@@ -41,4 +46,24 @@ class Meta:
             'id',
             'driver', 'car', 'passengers',
             'date', 'start_mileage', 'end_mileage', 'description',
+            'start_location', 'end_location',
         )
+
+    def create(self, validated_data):
+        # tracks_data = validated_data.pop('tracks')
```
weronikafelczak: Why is this line commented out?

weronikafelczak:
Please resolve conflict.

backend/fleet_management/serializers.py
jacol12345: Good spot :D I copied it from tutorial when I was trying to figure out how process nested data.

PR #86, opened 2018-12-01
Action creators

adamgajzlerowicz:
This PR adds endpoint consumer factory. It is fully tested - pulls data correctly and populates the form with car records.
Once merged further work will be carried out


frontend/src/views/RouteFormView.vue
```diff
@@ -163,10 +166,15 @@ export default {
     },
   },
   created() {
-    this[actions.FETCH_CARS]();
+    this.fetchDrives();
```
jacol12345: If we want to rely 100% on constants then you can change it to `this[actions.fetchCars]()`

frontend/src/views/RoutesView.vue
```diff
@@ -1,4 +1,5 @@
 <template>
+
```
weronikafelczak: Please remove empty line.

frontend/src/views/RoutesView.vue
```diff
@@ -44,7 +45,6 @@
     </div>
   </div>
 </template>
```
weronikafelczak: Please add empty line.

PR #89, opened 2018-12-03
Makefile cleanup

magul:
I removed `-docker` sufix form linting and testing command, as it works perfectly fine inside docker containers and we do not run these command outhide.

Also figure out how to attach to running Django instance. If somebody want to debug Django application, just put `import pdb; pdb.set_trace()` in Django code, and run `make debug-backend`.

To detach You need to press CTRL+`p` and CTRL+`q`.

Will still work on cleaning up nginx setup, but that will be done in separate pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,37 @@
 # pah-fm
 
 
-# Build project
+## Run project (in background)
 
 ```
-make build
+make start
 ```
 
-# Run project
-
+## Stop project
 ```
-docker-compose up
+make stop
 ```
 
-# Run project locally for testing
+## Attach to logs
 ```
-docker-compose up -d db
-
-virtualenv env -p /usr/local/bin/python3
-source env/bin/activate
-
-pip install -r backend/requirements/dev.txt
-
-make manage CMD=createsuperuser
+make logs
 ```
+
+## Run linters
+```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,29 +1,37 @@
...
+```
+make test
+```
 
-make manage CMD=createsuperuser
+## Debug backend (Django)
+```
+make debug-backend
 ```
+After finishing debugging detach from shell using *CTRL*+`p` and *CTRL*+`q`.
```
adamgajzlerowicz: Can we please also add section on accessing frontend and admin pages?

magul: As I wrote in description I have a plan to do nginx cleanup and prefer not to mix these two things.

BTW what does it mean?
> accessing frontend
 Do You want to attach to node process, that run webpack builder? What for?

adamgajzlerowicz: I don't know what this means. Where is this from?

magul: You just wrote about "accessing frontend and admin pages" - I assumed, that You want to access proccess that run in frontend docker container responsible for hot reload of Vue changes.

If You were refering just what to write in browser address bar, I would like to handle it with nginx cleanup (which I'm working on).

PR #90, opened 2018-12-03
Mobile navigation and re-skin

adamgajzlerowicz:
This PR brings

- Hamburger menu
- PAH logo (polish if polish language selected, english logo if any other language selected)
- Makes theme more unified with PAH website theme
- Fixes footer on mobile
- Fixes flag display


frontend/src/App.vue
```diff
@@ -31,25 +71,94 @@ import store from './store';
...
+  top: 1em;
+}
+
+.language-mobile {
+  display: flex;
+  margin-top: 50px;
+
+  & .lang {
+    margin: 0 auto;
+    padding: 0 !important;
```
jacol12345: Out of curiosity, which style overwrites that?

frontend/src/components/Header.vue
```diff
@@ -1,69 +1,152 @@
...
+    </a>
+    <nav class="menu">
+      <b-nav
+        fill
       >
-        {{ link.text }}
-      </b-nav-item>
-      <LoginStatus v-bind="user"/>
-    </b-nav>
+        <b-nav-item
```
jacol12345: Can menu be a separate component? Now it's duplicated in Header and App components.

frontend/src/components/Language.vue
```diff
@@ -7,7 +7,8 @@
         :key="language"
         @click="changeLang(language)">
         <flag
-          :iso="language" />
+          :iso="language"
+          :squared="false"/>
```
jacol12345: 💎

frontend/src/components/Header.vue
```diff
@@ -1,69 +1,152 @@
...
+      min-height: 100px;
+      font-size: 20px;
+      min-width: 155px;
+      color: $white;
+      display: flex;
+      align-items: flex-end;
+    }
+
+    & > li:nth-child(1) {
+      background: $pah-color-1;
```
jacol12345: We could have CSS classes with these colors and assign them in HTML.  I'm a fan of this solution but yours is equally good.

jacol12345:
A couple of styling polishing on small screen (tested on Chrome with iPhone5 emulation)

Menu overlapses logo:
screen shot 2018-12-05 at 07 42 17

Is it intended that not all items in menu are displayed one below another (Routes and Login are on the same line)
screen shot 2018-12-05 at 07 42 33

Routes list: items should take 100% of the available container width.
screen shot 2018-12-05 at 07 43 52

On desktop (900px width) menu goes too high and language switcher is displayed on top of it.
Maybe we could use grid for the menu on desktop?
screen shot 2018-12-05 at 07 47 18

adamgajzlerowicz:
This is navigation work not pages work of particular pages though. If they need to be restyles, let's make a separate task for it. 
I am not sure why you are testing page on IPhone 5 - the target audience is users on android devices. If other devices need to be considered though, please make a separate task that specifies those so they can be addressed properly. 

> A couple of styling polishing on small screen (tested on Chrome with iPhone5 emulation)
> 
> Menu overlapses logo:
> screen shot 2018-12-05 at 07 42 17
> 
> Is it intended that not all items in menu are displayed one below another (Routes and Login are on the same line)
> screen shot 2018-12-05 at 07 42 33
> 
> Routes list: items should take 100% of the available container width.
> screen shot 2018-12-05 at 07 43 52
> 
> On desktop (900px width) menu goes too high and language switcher is displayed on top of it.
> Maybe we could use grid for the menu on desktop?
> screen shot 2018-12-05 at 07 47 18

jacol12345:
I test with Chrome emulator set to iPhone 5 because it's a preset with a small screen. As far as I know it's the only difference between presets, so I could use an older Samsung Galaxy phone as well.

It has to work very good on mobile first, so I still think these tweaks are required.

PR #98, opened 2018-12-08
#65 Send emails

### LINKED ISSUE #65: Confirmation: send e-mail to passengers
AC:
* After new drive is created, an e-mail to each participant should be sent
* The e-mail contains messages (text below)
* The e-mail contains links to confirmation screen; the link is unique for passenger and drive and include language information (i.e. /confirm/<hash>/pl/)
* Clicking on link should show a page with basic drive data (rendered by Django)
* There should be a link below each language version, leading to the same-looking page but in the same language as message it is connected with

Message text:
> Thanks for driving with me on {{ date }} from {{from }} to {{ destination }}. Please confirm your participation and select your project by clicking on the link below. Sincerely, {{ driver username }}

The above message should be available in all languages (Polish, English and Ukrainian for now).
Message should be hard-coded. If any of the translations is not present just use Google Translate.

jacol12345:
Closes #65 

In order to test run backend tests, then inspect emails content in backend/emails directory.


backend/fleet_management/templates/new_drive_email.html
```diff
@@ -0,0 +1,28 @@
+Hello {{ passenger.first_name }}!
```
adamgajzlerowicz: How about adding a locale field to passenger and basing email content based on that?

backend/fleet_management/templates/new_drive_email.html
```diff
@@ -0,0 +1,28 @@
...
+
+LINK
+
+щиро, {{ driver.first_name }} {{ driver.last_name }}
+
+---------------------------------------------------------
+
+Witaj {{ passenger.first_name }}!
+Dziękuję za przejazd ze mną {{ drive.date }} z {{ drive.start_location }} do {{ drive.end_location }}.
+Potwierdź swój udzial i wybierz project klikając na link poniżej.
```
adamgajzlerowicz: Please add polish characters

backend/fleet_management/templates/new_drive_email.html
jacekbj: I thought about it but I have no strong opinion about that. @danutaj23 What do you think about this idea?

backend/fleet_management/templates/new_drive_email.html
jacekbj: I can't :D I have "option + l" mapped in my IDE. Can you turn a blind eye here :D ?

adamgajzlerowicz: That sounds good - after all, all translations should be reviewed by a non dev person.
Iam wondering though if we should put all of the translations into one common file placed in the root of the project, rather then having one for backend and one for frontned. Should we create a task for that? What do you think?

backend/fleet_management/templates/new_drive_email.html
adamgajzlerowicz: I am imagining being a user of the application who speaks primarily polish. I'd need to scroll to the bottom of the email on my phone each time when I get it and I'd get discouraged easily.

jacol12345:
I wonder how small your phone would have to be for such experience :D
Anyway, as it is is super-simple to test because it has lower complexity.

adamgajzlerowicz:
I was told that we go down to 320px. I believe you suggested checking on Iphone 5? Did you have a chance to test it on that?

jacol12345:
As per my notes, the only way to test it is to take a look at the logs.
Email service is not configured. An easy way would be to copy the rendered
email content into email client and just send it to myself :)

I've given it some thought and I am against adding the preferred language
atm. My thinking: we should focus on delivering end-to-end experience now.
With language pref we would need to sync language between backend and
frontend. Having one source of string templates would require adding
additional code to Django app whereas I imagine there will be only 1-2 more
templates needed there.

adamgajzlerowicz:
It really sounds like django is a very inflexible framework. Having backend in node it would be quicker to add this feature than having this conversation hah
Having translated texts in two locations feels like will required admin overhead, but its probably not my call

jacol12345:
I wouldn't say Django is inflexible. Sharing code between backend and frontend always comes at a price when these are not written in the same language.

backend/fleet_management/apps.py
```diff
@@ -3,3 +3,6 @@
 
 class FleetManagementConfig(AppConfig):
     name = 'fleet_management'
+
+    def ready(self):
+        from .signals import *
```
KarolJagodzinski: It's a bad practice to import outside `imports section` on the top of the file. and please don't import `*`.
You may go with `from . import signals` or `from .signals import What, You, Need`

backend/fleet_management/migrations/0005_passenger_email.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('fleet_management', '0004_auto_20181129_2056'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='passenger',
+            name='email',
+            field=models.EmailField(default='email@email.com', max_length=254),
```
KarolJagodzinski: I guess `email@email.com` is an example data not a valid value which should be in database. If I'm right please change that.

backend/fleet_management/serializers.py
```diff
@@ -41,4 +47,30 @@ class Meta:
             'id',
             'driver', 'car', 'passengers',
             'date', 'start_mileage', 'end_mileage', 'description',
+            'start_location', 'end_location',
         )
+
+    def create(self, validated_data):
+        # tracks_data = validated_data.pop('tracks')
```
KarolJagodzinski: Please remove if not necessary

backend/fleet_management/serializers.py
```diff
@@ -41,4 +47,30 @@ class Meta:
...
+        passengers_data = validated_data.pop('passengers')
+        car_data = validated_data.pop('car')
+        car = Car.objects.get(pk=car_data['id'])
+        passengers = Passenger.objects.filter(
+            id__in=[p['id'] for p in passengers_data],
```
KarolJagodzinski: instead of `p` I'll suggest to pass `passanger_data` is much more readable than `p`

backend/fleet_management/serializers.py
```diff
@@ -41,4 +47,30 @@ class Meta:
...
+                driver=self.context['driver'],
+                car=car,
+            )
+            drive.passengers.set(passengers)
+            drive.save()
+            drive_created.send(
+                self.__class__,
+                drive_id=drive.id,
+                driver_id=self.context['driver'].id,
+                passengers_ids=[p.id for p in passengers],
```
KarolJagodzinski: ditto

backend/fleet_management/signals.py
```diff
@@ -0,0 +1,38 @@
...
+            address = passenger.email
+            email = EmailMessage(
+                subject=subject,
+                body=message,
+                from_email=EMAIL_ADDRESS,
+                to=[address],
+                connection=connection,
+            )
+            email.send()
+            email.content_subtype = 'html'
```
KarolJagodzinski: Shouldn't it be above `email.send()` ?

backend/fleet_management/test_drives.py
```diff
@@ -117,3 +125,40 @@ def test_can_retrieve_only_my_drives(self):
         drives = json.loads(res.content)
         self.assertEqual(len(drives), 1)
         self.assertEqual(drives[0]['id'], self.drives[0].id)
+
+    # this test allows to verify email template
```
KarolJagodzinski: comments should be inside a function / method

backend/fleet_management/test_drives.py
```diff
@@ -117,3 +125,40 @@ def test_can_retrieve_only_my_drives(self):
...
+        }
+
+        self.client.force_login(self.driver)
+        res = self.client.post(self.url, data=payload, format='json')
+
+        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
+
+        drive = Drive.objects.get(pk=res.data['id'])
+        self.assertSetEqual(
+            {p.id for p in drive.passengers.all()},
```
KarolJagodzinski: ditto

backend/fleet_management/test_drives.py
```diff
@@ -117,3 +125,40 @@ def test_can_retrieve_only_my_drives(self):
...
+            'description': '',
+            'startLocation': 'Warsaw',
+            'endLocation': 'Poznan',
+        }
+
+        self.client.force_login(self.driver)
+        res = self.client.post(self.url, data=payload, format='json')
+
+        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
+
```
KarolJagodzinski: you may assign `res.data` to some variable and reuse that

backend/fleet_management/apps.py
jacol12345: https://docs.djangoproject.com/pl/2.1/topics/signals/
>  If you’re using the receiver() decorator, simply import the signals submodule inside ready().

KarolJagodzinski: nice, thx

Issue #74, opened 2018-11-25
Drives: form: validate mileage


jacol12345:
AC:
* There is validation checking that `endMileage > startMileage`.
* Form cannot be saved if validation does not pass.

weronikafelczak:
Closed with #99.

PR #99, opened 2018-12-08
#74 Form validation

adamgajzlerowicz:


frontend/src/views/RouteFormView.vue
```diff
@@ -147,21 +144,32 @@ export default {
...
+        .reduce(makeErrorsPartial, {});
+
+      const { startMileage, endMileage } = this.route;
+
+      if (
+        !!startMileage
+        && !!endMileage
+        && parseInt(startMileage, 10) >= parseInt(endMileage, 10)
+      ) {
+        this.errors.startMileage = this.$t('common.start_mileage_error');
```
jacol12345: Errors are not translated if I switch language after validating form. We can live with that

jacol12345: There should be a separate error for a case when startMileage >= endMileage. It is confusing to see messages "Start mileage is incorrect" and "End mileage is incorrect" in such case.

frontend/src/views/services.js
```diff
@@ -1,13 +1,23 @@
 const requiredFields = [
-  'date', 'car', 'description', 'from', 'destination', 'startMileage', 'endMileage',
+  'date',
+  'car',
+  'description',
```
jacol12345: A remotely related thing: text fields are validated as correct when filled with spaces. This one we need to fix.

frontend/src/views/RouteFormView.vue
adamgajzlerowicz: I did consider that, but making error message that is way too specific might be just hard to read and confusing. Let's just make sure that the values are correct and user knows about incorrect ones.

Issue #96, opened 2018-12-06
link behind PAH's logo should be directed to the application's home page


monikakrawczak:
PAH's logo should be directed to the application's home page not to PAH's website.
Redirection to the PAH's can be next to Code For Poznań link at the bottom of the website.

weronikafelczak:
Closed with #100

PR #100, opened 2018-12-08
#96 moved pah link to footer

adamgajzlerowicz:


weronikafelczak:
@jacol12345 I think we can make a separate issue for that. :)

Issue #97, opened 2018-12-08
Put application behind login route


adamgajzlerowicz:
As there is no point accessing the form when user is not logged in, application should be secured with login
AC:
- Upon entering application if no session data found redirect to login
- Login page is the only available page
- Hide navigation on login
- After login perform initial data caching

weronikafelczak:
Closed with #101.

PR #101, opened 2018-12-09
#97 App behind login

adamgajzlerowicz:


frontend/src/components/Status.vue
```diff
@@ -1,16 +1,24 @@
...
+import { LOGIN_PATH } from '../router/constants';
 
 export default {
   name: 'Refresh',
   computed: {
     ...mapState([IS_ONLINE]),
   },
+  data() {
+    return {
+      isLogin: this.$router.currentRoute.path === LOGIN_PATH,
```
jacol12345: Is there any way in Vue to extract setting this property and share it between components? It is repeated a couple of times.

frontend/src/components/Status.vue
```diff
@@ -31,5 +39,9 @@ export default {
 .online {
   background-color: $green;
 }
+
+.hidden {
```
jacol12345: This class should be placed in "common" CSS file.

frontend/src/components/Status.vue
adamgajzlerowicz: I can see where you are coming from, thought I am trying to get away from monolithic architecture with the main view class holding entire knowledge for the entire app. Thus it is not bound to the `vue` object. 
I could extract it to the helper function, but it would just compare two strings, so I decided to leave it as it is

frontend/src/components/Status.vue
adamgajzlerowicz: Totally agree. I removed the class name entirely and used the same strategy as in the other places (conditional rendering based on props)

PR #111, opened 2018-12-13
fix login redirect

adamgajzlerowicz:


frontend/src/store/actions.js
```diff
@@ -22,8 +25,7 @@ export const actions = {
       .then((token) => {
         commit(mutations.SET_LOGIN_ERROR, null);
         saveToken(token);
-        dispatch(FETCH_USER);
-        window.location.replace('/');
+        dispatch(FETCH_USER, { callback: () => window.location.replace('/') });
```
jacol12345: Do we call `FETCH_USER` action anywhere else? If no, than instead of passing callback we can just redirect inside action.

adamgajzlerowicz: That would couple logic to implementation and break open-closed principle

PR #112, opened 2018-12-13
login view testcase

monikakrawczak:


monikakrawczak:
@w1stler @danutaj23 please review :)

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
+# Login testcase
+
+## Prerequisites
+User has acount in the system already created by system admin.
```
w1stler: 'acount' -> 'account'

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
+# Login testcase
```
w1stler: I would add information, that this is positive scenario (passing valid login and password). You can create at least one more scenario (with not-valid data + checking the validation of the fields).

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
...
+User has acount in the system already created by system admin.
+
+## Steps
+
+### Step 1
+User opens pah-fm website.
+
+### Expected results 1
+User sees login view with 2 text fields: "Username" and "Password" and 1 button "Login".
+User sees 3 flages: polish, english, ukrainian, in the right side, at top of the page.
```
w1stler: 'flages' -> 'flags'

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Step 2
+User enters his email (attached to already created account) into "Username" field.
+User enters his password (attached to already created account) into "Password" field.
+User clicks "Login" button.
+
+### Expected results 2
+User succesfully logs in and sees main view of application.
```
w1stler: Instead of main view I would say - is redirected to homepage (`/`). "Main view" might be misleading.

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
+# Login testcase
+
+## Prerequisites
+User has acount in the system already created by system admin.
+
+## Steps
+
+### Step 1
+User opens pah-fm website.
```
w1stler: At some point we would need to distinguish instance address (eg. development, release candidate or production) + that we would need to have example accounts there. As a rule of thumb there are the same accounts on every single one of them aside from the passwords - that allows testers/business people to easily test something out. 

Treat it more as a general comment.

docs/login-testcase.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Expected results 1
+User sees login view with 2 text fields: "Username" and "Password" and 1 button "Login".
+User sees 3 flages: polish, english, ukrainian, in the right side, at top of the page.
+
+### Step 2
+User enters his email (attached to already created account) into "Username" field.
```
w1stler: We might consider using `their` as possessive pronoun (a neutral one).

features/login-testcase.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Login testcase.
```
w1stler: Avoid using periods at the end.

features/login-testcase.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Login testcase. 
+
+    Positive scenario (passing valid login and password).
```
w1stler: This has to be posted as a comment (it would crash when you will try to run it). 

So any stuff that is not Gherkin-related, have to be marked as comment. In Python we use "#" for single line comment or """...""" for multiline comment. Please have a look here: https://www.pythonforbeginners.com/comments/comments-in-python

Example:

`# one line comment`

```
"""This
is
multi
line
comment.
"""```

File: features/login-testcase.feature [hunk 9]
```diff
@@ -0,0 +1,13 @@
...
+
+    Scenario: User opens pah-fm website.
+       Given User has account in the system already created by system admin.
```
w1stler: I am mostly familiar with the first-person-perspective (Given I am on X website, When I click something, Then the result is visible), although this is not wrong. Please be aware, that "I" notation is most commonly used.

features/login-testcase.feature
```diff
@@ -0,0 +1,13 @@
...
+        When User opens pah-fm website
+        Then User sees login view with 2 text fields: "Username" and "Password" and 1 button "Login".
```
w1stler: This is also bad habit - using mixed whens and thens. Please try to split the scenarios for small ones, where we have structure:

```
Given
When
Then
```

not

```
Given
When
Then
When
Then
```

w1stler: I am aware that this might not be clear at this point, although for the case of the automated tests - try to make them short and test one thing.

monikakrawczak:
@w1stler please review once again

features/login-testcase.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Login view
+
+    Scenario: User opens pah-fm login view
+        When User opens pah-fm website
+        Then User sees login view with 2 text fields: "Username" and "Password" and 1 button "Login"
```
w1stler: If there are number of elements, you could use table inside the step. Not needed here though. 

```
Scenario: some scenario
  Given a set of specific users
     | name      | department  |
     | Barry     | Beer Cans   |
     | Pudey     | Silly Walks |
     | Two-Lumps | Silly Walks |

 When we count the number of people in each department
 Then we will find two people in "Silly Walks"
  But we will find one person in "Beer Cans"
```

PR #113, opened 2018-12-15
add new route testcase

monikakrawczak:
@w1stler @danutaj23 please review :)


docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
+# Add New Route testcase
+
+ ## Prerequisites
+
+User has acount in the system already created by system admin.
```
w1stler: ditto (account)

docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
...
+User is log in.
```
w1stler: Is logged in (passive voice)

docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
+# Add New Route testcase
+
+ ## Prerequisites
+
+User has acount in the system already created by system admin.
+User is log in. 
+User is on a main view.
```
w1stler: I would also add direct link - "main view" might be misleading (and main view might differ for different users in the future).

docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
...
+
+ ## Steps
+
+ ### Step 1
+User clicks "Add New Route" button at the top of the page.
+If User use screen with lower resolution "Add New Route" button is hidden in hamburger menu.
```
w1stler: We should think of separate scenarios for mobile views. It would be more clean in my opinion. As we would not strict breakpoints (or I am not sure exactly, how Vue will handle that), I would go with separate scenarios for every view with RWD tag (just to ensure, that the test / happy path is performed on mobile device).

docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
...
+1. "Starting Mileage" - number field
+1. "Ending Mileage" - number field
+and 1 button "Submit" at the bottom of form.
+
+ ### Step 2
+User enters correct data
+User clicks "Submit" button.
+
+ ### Expected results 2
+User is informed about succes and redirected to routes list view.
```
w1stler: 'succes' -> 'success'

redirect - I would add link

docs/add-new-route-testcase.md
```diff
@@ -0,0 +1,31 @@
...
+1. "Choose a car" - dropdown field
+1. "Description" - text field
+1. "From" - text field
+1. "Destination" - text field
+1. "Starting Mileage" - number field
+1. "Ending Mileage" - number field
+and 1 button "Submit" at the bottom of form.
+
+ ### Step 2
+User enters correct data
```
w1stler: You can think of validation scenario (validation for every field, with max./min. values, etc.).

w1stler: Also: "correct data" is too loose for me - you can add examples in commas for example, which might be helpful for user that would read that. These are self-explanatory, although you've already raised an issue, that you can add special characters to the milage fields (#114). That would look:

1. Date - date field, eg. 11/12/2018 (mask does matter here - order of writing the date). ISO 8601 states how the date should look like, FYI. 
1. Choose a car - dropdown field (information, that this is single-select only, not multiple values could be selected)
1. Description - text field (eg. "Lorem ipsum et dolores")

monikakrawczak:
@w1stler @danutaj23 @yellowee please review

features/add-new-route-testcase.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Add New Route
+
+    Scenario: User opens pah-fm Add New Route view
+       Given User has account in the system already created by system admin
+         And User is already logged in
```
w1stler: This two steps might be consolidated - Given user is logged in (which implies, that the account is already created).

features/add-new-route-testcase.feature
```diff
@@ -0,0 +1,32 @@
...
+        When User opens "/routes/" view
+        Then User sees "Date" - date field dd.mm.rrrr
```
w1stler: Table would be useful here (field - first column, data type - second field).

features/add-new-route-testcase.feature
```diff
@@ -0,0 +1,32 @@
...
+         And User sees "From" - text field
+         And User sees "Destination" - text field
+         And User sees "Starting Mileage" - number field
+         And User sees "Ending Mileage" - number field
+         And User sees button "Submit"
+
+    # Positive scenario for large displays
+    Scenario: User adds new route
+       Given User has account in the system already created by system admin
+         And User is already logged in
```
w1stler: Ditto

PR #117, opened 2018-12-17
Logout function

lukaszmaterla:


frontend/src/components/NavigationItems.vue
```diff
@@ -12,32 +12,31 @@
...
 
 export default {
   name: 'NavigationItems',
   computed: {
     ...mapState(['user']),
   },
+  methods: {
+    ...mapActions([actions.LOGOUT]),
+    logout() {
+      this[actions.LOGOUT]();
```
adamgajzlerowicz: this seems to be here twice now

frontend/src/components/Status.vue
```diff
@@ -14,6 +18,7 @@ export default {
   name: 'Refresh',
   computed: {
     ...mapState([IS_ONLINE]),
+    ...mapState(['user']),
```
adamgajzlerowicz: I believe that those two lines can be joined into one.

adamgajzlerowicz:
Can we actually add the `/logout` path to the router (`src/router/index.js`)?  Currently visiting /logout doesn't trigger the logout action, but redirects to `/`

lukaszmaterla:
I'm not sure, but this is because of beforeEach method in index.js.
Do we actually need this route? I tried to change this, and execute logout actions from store but it didn't work in beforeEnter hook.

adamgajzlerowicz:
For this a new route is required in `src/router/index.js`. This route will have a component that will be dedicated to perform a logout and redirect back to login after the logout. I hope that this is making sense. PM if it doesn't 😄

jacol12345:
Merged as part of #138

PR #123, opened 2018-12-28
Changed layout rendering

adamgajzlerowicz:
This changes main layout rendering strategy to utilise space-between provided by flex. Benefit of this solution is that changes to layout will be easier to do in future. Also calc is removed so hopefully any potential bugs in browsers will be nullified.

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -87,6 +87,10 @@ export default {
   background: $white;
 }
 
+html {
+  height: auto;
+}
+
 #page-wrap {
   margin-bottom: auto;
 }
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -12,7 +12,7 @@
     </ScaleRotate>
     <div
       id="page-wrap"
-      class="container page-container page-wrap">
+      class="container">
       <div class="row">
         <div class="col-sm-12">
           <transition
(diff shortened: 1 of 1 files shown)


jacol12345:
There is an issue with vh on Chrome:
https://dev.to/peiche/100vh-behavior-on-chrome-2hm8
this results in Footer not sticked to the bottom of the page.

I managed to overcome this with the following code:

```
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -79,12 +79,13 @@ export default {

 #app {
   @include flex($direction: column, $justify: space-between);
-  min-height: 100vh;
+  min-height: 100%;
   font-family: 'Avenir', Helvetica, Arial, sans-serif;

 #page-wrap {
diff --git a/frontend/src/components/Footer.vue b/frontend/src/components/Footer.vue
index b402085..6255d10 100644
--- a/frontend/src/components/Footer.vue
+++ b/frontend/src/components/Footer.vue
@@ -65,3 +65,9 @@ export default {
   bottom: 0;
 }
 </style>
+
+<style>
+    body {
+        height: 100%;
+    }
+</style>
:
```

It uses percentages instead of vh.

adamgajzlerowicz:
I have applied a fix, but I still need to test it with real device.

adamgajzlerowicz:
You are right - this just won't work without hacks and walkarounds, so I am closing the pr as the previous version just worked. Thanks for your help!

Issue #120, opened 2018-12-21
Drive: form: add start/end location fields


jacol12345:
Add start and end location fields.
Lookup API field names in backend code (serializers.py or even better check tests).
Min length: if not defined in Django model, then 3 characters.
Max length: if not defined in Django model, then 200 characters.

Show new fields on Drive list view.

PR #124, opened 2018-12-28
#120 Updated field names to correspond with api fields

adamgajzlerowicz:


adamgajzlerowicz:
Updated as per request

PR #131, opened 2019-01-16
Set min and max in start/end mileage.

### LINKED ISSUE #118: Set min and max attributes on mileage fields.
Set min=0 and max = 1.5e6 on start and end mileage fields. This will limit the range one can input into these fields.

jacol12345:
Closes #118

### DIFF
--- a/frontend/src/views/RouteFormView.vue
+++ b/frontend/src/views/RouteFormView.vue
@@ -92,6 +92,8 @@
                 <div class="form-group col-sm-6">
                   <label>{{ $t('routes.starting_mileage') }}</label>
                   <input
+                    min="0"
+                    max="1500000"
                     type="number"
                     v-model="route.startMileage"
                     name="startMileage"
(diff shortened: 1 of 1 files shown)


frontend/src/views/RouteFormView.vue
```diff
@@ -92,6 +92,8 @@
                 <div class="form-group col-sm-6">
                   <label>{{ $t('routes.starting_mileage') }}</label>
                   <input
+                    min="0"
+                    max="1500000"
```
adamgajzlerowicz: Maybe they have a reeeealy good diesel engines and they can go further than that? 😃

OtisRed: Due to curiosity I checked some ridiculously long routes and e.g. there is 13 632 km from Porto to Wladywostok :P In 1500000 you could circle the earth almost 38 times ^^

PR #139, opened 2019-01-18
Added fetch projects list and display list on confirmation page

lukaszmaterla:


frontend/src/views/ConfirmationView.vue
```diff
@@ -110,14 +133,19 @@ export default {
...
   created() {
     this[actions.VERIFY_CONFIRMATION_TOKEN](this.token);
+    this[apiActions.fetchProjects]();
   },
   computed: {
     ...mapState([VERIFICATION_TOKEN]),
     token() {
       return this.$route.params.token;
     },
+    ...mapState(namespaces.projects, {
```
jacol12345: Can you put that in line 139?
```
...mapState([VERIFICATION_TOKEN, namespaces.projects]),
```
I don't know if that works though :D

frontend/src/views/ConfirmationView.vue
```diff
@@ -110,14 +133,19 @@ export default {
...
+      projects: state => state,
+    }),
   },
   methods: {
+    ...mapActions(namespaces.projects, [apiActions.fetchProjects]),
```
jacol12345: ditto as for state: I think this could go onto array right below

frontend/src/views/ConfirmationView.vue
```diff
@@ -129,6 +157,7 @@ export default {
       const payload = {
         isOk: this.confirmation.ok === 'yes',
         comment: this.confirmation.comment,
+        project: this.confirmation.projects,
```
jacol12345: Here it's `this.confirmation.projects` but in line 174 it's `this.confirmation.project` (no `s`).

lukaszmaterla:
Hi, I am sick, but ok, I gonna do this :)

PR #140, opened 2019-01-20
logout testcase

yellowee:


yellowee:
@w1stler @monikakrawczak @danutaj23

features/logout-testcase.feature
```diff
@@ -0,0 +1,13 @@
...
+    Given User is logged into PAH website
+    When User clicks on Logout button
+    Then User's session details are removed
+    And User is redirected to the login screen
+
+Scenario: User logs out from PAH website using URL
+    Given User is logged into PAH website
+    When User enters URI '/logout' directly in the address bar
+    Then User's session details are removed
+    And User is redirected to the login screen
```
w1stler: Please add newline at the end.

features/logout-testcase.feature
```diff
@@ -0,0 +1,13 @@
+Feature: logout
```
w1stler: In general looks great!  

Please adjust formatting as @monikakrawczak did. You can also work out together the way of formatting (using minuscules and majuscules, way of naming features and scenarios, etc.). This is just to be consistent in every .feature file. 

Please ping me on gitter, if you have any further questions!

features/logout-testcase.feature
yellowee: +1 thanks

features/logout-testcase.feature
yellowee: I've adjusted the formatting and capitalization to match Monika's testcases

w1stler:
I would do that this way:

```Feature: test feature name

Scenario: User logs out from pah-fm website using Logout button
  Given User is logged into pah-fm website
   When User clicks on Logout button
   Then User's session details are removed
    And User is redirected to the Login view```

**Comment by magul** [comment]
there's something obviously wrong with indentation here (maybe You're using tabs instead of spaces here @yellowee? we should definitely use only spaces here).

As I was working on Gherkin files with @monikakrawczak We decided, that spaces inside steps will be justified to `:` in the scenario name.

Like this:
```
    Scenario: User opens pah-fm login view
        When User opens pah-fm website
```
We're open to suggestions (maybe space in scenario and space in step?), but definately indentations in here do not look well.

yellowee:
Hi, thank you for your comments. I was using 4 spaces, but I checked with Gherkin documentation and they are using 2 spaces. 
If it's easier to read - I've reindented the code to `:` as suggested

Issue #114, opened 2018-12-15
Mileage fields accept non-digits


monikakrawczak:
On "Add New Route" view fields "Starting Mileage" and "Ending Mileage" accept non-digits (`+`, `-`, `,` and `.`). 
Brower based validation works properly after clicking "Submit" button.

I'm aware that `,` and `.` could be used but `+` and `-` doesn't make sense here.

Tested on Chrome.

adamgajzlerowicz:
https://github.com/CodeForPoznan/pah-fm/pull/163

PR #163, opened 2019-02-04
Bug/114 input numbers only

adamgajzlerowicz:
https://github.com/CodeForPoznan/pah-fm/issues/114

If you can make it work any other way please do tell me...

### DIFF
--- a/frontend/src/views/RouteFormView.vue
+++ b/frontend/src/views/RouteFormView.vue
@@ -95,7 +95,8 @@
                   <label>{{ $t('routes.starting_mileage') }}</label>
                   <input
                     min="0"
-                    max="1500000"
+                    onkeypress="return event.charCode >= 48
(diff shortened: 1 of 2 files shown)


frontend/src/views/RouteFormView.vue
```diff
@@ -95,7 +95,10 @@
                   <label>{{ $t('routes.starting_mileage') }}</label>
                   <input
                     min="0"
-                    max="1500000"
+                    onkeypress="return event.key === 'Enter'
+                        || (Number(event.key) >= 0
+                        && Number(event.key) <= 9
```
juliaanholcer: What do you think about using `!isNaN(event.target.value)` instead of two conditions with `Number(event.key)` ?

adamgajzlerowicz: Unfortunately that still evaluates to number if characters like `e`, `.`, or `+` are passed

PR #166, opened 2019-02-13
login testcase in behave

monikakrawczak:


features/steps/steps.py
```diff
@@ -0,0 +1,24 @@
...
+driver = webdriver.Chrome('./chromedriver')
+
+@when(u'User opens pah-fm website')
+def step_impl(context):
+    driver.get('http://localhost:8080/login')
+
+
+@then(u'User sees login view with two text fields: "Username" and "Password" and one button "Login"')
+def step_impl(context):
+    time.sleep(1)
```
w1stler: Read about expected conditions. TL;dr - using time.sleep is an anti-pattern. ;D

https://www.seleniumhq.org/docs/04_webdriver_advanced.jsp

PR #168, opened 2019-02-19
Full login features in behave

monikakrawczak:
@w1stler @yellowee 
please review :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,3 +68,6 @@ deploy-frontend-heroku:
 
 send-test-email-heroku:
 	make manage-heroku CMD="send_test_mail ${EMAIL}"
+
+init-behave:
+	cat features/data/fixtures.sql| docker-compose exec -T db psql -U pah-fm
--- a/features/data/fixtures.sql
+++ b/features/data/fixtures.sql
@@ -1,0 +1,3 @@
+INSERT INTO public.fleet_management_user(
+    password, is_superuser, username, first_name, last_name, email, is_staff, is_active, date_joined)
+    VALUES ('pbkdf2_sha256$120000$B4VUk7IZMZCk$t9nZBMo6q4EghFgeqkTLhstGRTlYCRfHJtGkuNzUTK0=', FALSE, 'solomiiavyshniak@eukr.net', 'Ада', 'Гавриленко', 'solomiiavyshniak@ukr.net', FALSE, TRUE, '2019-02-18');
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,0 +1,8 @@
+from selenium import webdriver
+
+def before_all(context):
+    context.driver = webdriver.Chrome('./chromedriver')
+
+
+def after_all(context):
+    context.driver.quit()
--- a/features/steps/steps.py
+++ b/features/steps/steps.py
@@ -1,24 +1,51 @@
-from selenium import webdriver
 from selenium.webdriver.common.keys import Keys
 import time
 
-driver = webdriver.Chrome('./chromedriver')
+
 
 @when(u'User opens pah-fm website')
 def step_impl(context):
-    driver.get('http://localhost:8080/login')
+    context.driver.get('http://localhost:8080/login')
 
 
-@then(u'User sees login view with two text fields: "Username" and "Password" and one button "Login"')
-def step_impl(context):
-    time.sleep(1)
-    driver.find_element_by_name('username')
-    driver.find_element_by_name('password')
-    driver.find_element_by_css_selector('button.btn.btn-primary')
+@then(u'User sees "{label}" {type_} field')
+def step_impl(context, label, type_):
+    input_element = context.driver.find_element_by_xpath("//label[text()='" + label + "']/../input")
+    assert input_element.get_attribute('type') == type_
+
+@then(u'User sees "{text}" button')
+def step_impl(context, text):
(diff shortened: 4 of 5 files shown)


Makefile
```diff
@@ -68,3 +68,6 @@ deploy-frontend-heroku:
 
 send-test-email-heroku:
 	make manage-heroku CMD="send_test_mail ${EMAIL}"
+
+init-behave:
+	cat features/data/fixtures.sql| docker-compose exec -T db psql -U pah-fm
```
w1stler: Space before pipe (`|`)?

features/data/fixtures.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO public.fleet_management_user(
+    password, is_superuser, username, first_name, last_name, email, is_staff, is_active, date_joined)
+    VALUES ('pbkdf2_sha256$120000$B4VUk7IZMZCk$t9nZBMo6q4EghFgeqkTLhstGRTlYCRfHJtGkuNzUTK0=', FALSE, 'solomiiavyshniak@eukr.net', 'Ада', 'Гавриленко', 'solomiiavyshniak@ukr.net', FALSE, TRUE, '2019-02-18');
```
w1stler: I guess we would also need to user fixtures in other scenarios. We can think of a better way to create model users before testing whole app / runninig them locally.

I would rather not use sql statements but. Plus there is (probably) real email, so we should avoid that. Let's create model users which could be used in behave, unit and manual tests as well (shared dict or something like that).

Also no newline at the end of the file.

features/environment.py
```diff
@@ -0,0 +1,8 @@
+from selenium import webdriver
+
+def before_all(context):
+    context.driver = webdriver.Chrome('./chromedriver')
```
w1stler: Will that work with Chromedriver in default PATH `/usr/local/bin`?

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
+
 
 @when(u'User opens pah-fm website')
 def step_impl(context):
-    driver.get('http://localhost:8080/login')
+    context.driver.get('http://localhost:8080/login')
+
 
+@then(u'User sees "{label}" {type_} field')
+def step_impl(context, label, type_):
```
w1stler: Please add docstrings.

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
+
+@then(u'User sees three flags: polish, english, ukrainian, in the right side, at top of the page')
+def step_impl(context):
+    context.driver.find_element_by_css_selector('span.flag-icon.flag-icon-pl')
+    context.driver.find_element_by_css_selector('span.flag-icon.flag-icon-gb')
+    context.driver.find_element_by_css_selector('span.flag-icon.flag-icon-ua')
+
+
+
+
```
w1stler: Too many blank lines.

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
 @when(u'User opens pah-fm website')
 def step_impl(context):
-    driver.get('http://localhost:8080/login')
+    context.driver.get('http://localhost:8080/login')
+
 
+@then(u'User sees "{label}" {type_} field')
+def step_impl(context, label, type_):
+    input_element = context.driver.find_element_by_xpath("//label[text()='" + label + "']/../input")
+    assert input_element.get_attribute('type') == type_
```
w1stler: Can we add some messages into assert functions?

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
+def step_impl(context, text):
+    context.driver.find_element_by_xpath("//button[text()='" + text + "']").click()
+
+@then(u'User succesfully logs in')
 def step_impl(context):
     time.sleep(1)
-    driver.find_element_by_name('username')
-    driver.find_element_by_name('password')
-    driver.find_element_by_css_selector('button.btn.btn-primary')
+    assert context.driver.execute_script("return window.localStorage.jwt") is not None
```
w1stler: Ditto

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
 
-@then(u'User sees three flags: polish, english, ukrainian, in the right side, at top of the page')
+@then(u'User is redirected to homepage')
 def step_impl(context):
-    driver.find_element_by_css_selector('span.flag-icon.flag-icon-pl')
-    driver.find_element_by_css_selector('span.flag-icon.flag-icon-gb')
-    driver.find_element_by_css_selector('span.flag-icon.flag-icon-ua')
-    driver.quit()
+    time.sleep(1)
```
w1stler: Please try to add condition that the webpage is opened (to avoid using time.sleep, which is not a good pattern).

features/steps/steps.py
```diff
@@ -1,24 +1,51 @@
...
+    assert context.driver.current_url == 'http://localhost:8080/'
```
w1stler: FYI, this will fail when running on external instance. ;-)

jacol12345:
I am getting an error:
`ConfigError: No steps directory in '/Users/${User}/pah-fm/pah_behave/features'.

w1stler:
I've adjusted this PR on my branch (it was easier for me). I've pulled your commit into mine and made proper changes. ;-)

PR #176, opened 2019-03-01
Removed required from description

adamgajzlerowicz:


frontend/src/views/RoutesView.vue
```diff
@@ -114,7 +114,7 @@ export default {
   computed: {
     ...mapState(['routes']),
     ...mapState(namespaces.drives, {
-      drives: state => state.data || [],
+      drives: state => [...state.data].reverse(),
```
adamgajzlerowicz: puts latest drive on top

PR #177, opened 2019-03-02
Fix interface bugs

adamgajzlerowicz:
This fixes all interface bugs - tested on a real device

before: Username next to the online dot was overlaping top logo
now: username moved to sidebar next to logout link

before: no posiblity to change language on login screen. On a tablet size screen flags visible but not clickabe as of overlapping login form
now: moved flags to under login screen

before: Footer broken on android devices, which was depending on keyboard open/closed
now: Footer removed and CFP identification moved to sidebar

before: after login navuigate to `/`
now: after login navigate to drive form

before: Drive form not taking entire screen width
now: Drive form taking most of space on the screen

before: Different padding on different pages
now: All paddings are equal

before: Weird scroll on drives, which was very narrow and unreadable
now: Drives scrolls as normal

before: header had different size on mobile desktop
now: all desktop navigation removed, and special logic that was used to supported desktop is removed. Also application has now only one menu.

Perhaps those tweaks are not perfect, but feels to me that they are removing obvious bugs we had in the app.


yellowee:
Manual tests ok:
1. login-testcase written previously in Gherkin still applies, **but the flags are visible at the bottom of the screen**[!] (testcase must be updated)
2. logout-testcase written previously in Gherkin still passes ✓
3. add-new-route written previously in Gherkin still passes **with the exception of the redirecting to the routes view** [!] (this feature might be added later IMO + testcase should be updated)
- **I was unable to add a route with a blank Description field** [!]
- **I managed to add a future-dated route** [!]
- I was unable to add a route with Ending Mileage smaller than Starting Mileage ✓
- I was unable to add a route with Ending Mileage the same as Starting Mileage ✓
- I was unable to put something different than a number in Ending Mileage and Starting Mileage fields ✓ 
- Routes view is sorted by the date the drive took place (from the oldest to the newest), not sure if this should be sorted by the date of the drive or the date the drive has been added
- Viewing a selected previous drive in the Routes view does not include all fields (missing: Start location, End location, not sure if Passengers should be also visible) [!]
4. language-testcases written previously in Gherkin might pass but all the form's fields have not been translated
5. Other manual tests performed:
- Unsynced drives added while offline automatically sync when online ✓ 👯‍♂️ 👍 
- What should happen when trying to login while offline? [!]

PR #183, opened 2019-03-06
drive model and serializer adjusted to work with project

konradgalczynski07:


backend/fleet_management/migrations/0008_auto_20190306_1810.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.1.2 on 2019-03-06 18:10
```
magul: please squash migrations into one

jacol12345:
Looks good. I'd love to see a unit tests to know that it actually works but if time is of the essence then it's 👍 from me.

adamgajzlerowicz:
Good work!

PR #206, opened 2019-03-13
pahfm.codeforpoznan.pl as Django ALLOWED_HOST

magul:


backend/pah_fm/settings.py
```diff
@@ -38,6 +38,7 @@
         'localhost',
         '127.0.0.1',
         '52.232.62.212',
```
adamgajzlerowicz: is this still required?

magul: most probably not, but I'm going to propose another pull request, that will clean all heroku residues (also with that hosts here).

PR #229, opened 2019-05-08
Created country model

werginia:
Created just country model


backend/fleet_management/admin.py
```diff
@@ -1,14 +1,16 @@
...
+from .models import Car, Passenger, Drive, User, Project, VerificationToken, Country
 
 
 admin.site.register(Car)
 admin.site.register(Passenger)
 admin.site.register(Drive)
 admin.site.register(User, UserAdmin)
 admin.site.register(Project)
+admin.site.register(Country)
+
```
jacekbj: The empty line here is not necessary. It's not harmful, just aesthetics :)

backend/fleet_management/migrations/0008_country.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.1.2 on 2019-05-08 18:24
+from __future__ import  unicode_literals
```
jacekbj: Just FYI, we're coding in Python 3 so this line is not required.
This is an auto-generated file and it's no harmful so there is no need to change it.

jacekbj:
@werginia Did you consider using an app for that purpose, like https://github.com/SmileyChris/django-countries? If yes, why did you decide not to follow that path? It has so many goodies it's hard to resist 🍰 🍷 🏖

backend/fleet_management/migrations/0008_country.py
magul: As is says line above, that's usually autogenerated by Django. It's difficult to say why Django is still adding that line here (as Django 2+ is Python 3 exclusive) but don't think, that changing auto-generated migration would make a lof of sense here.

PR #230, opened 2019-05-08
add total mileage sum of drives and unsynced drives to drives list

### LINKED ISSUE #212: total of kilometers done for all the trips
Reported by PAH:
> When clicking on 3 horyzontal lines on the right-top corner, and click to see the registered trips, can it as well somehow show the total of kilometers done for all the trips? Is it possible even?

juliaanholcer:
resolves #212

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -6,10 +6,11 @@ import { actions } from './actions';
 import { VERIFICATION_TOKEN, SYNC, namespaces, IS_ONLINE } from './constants';
 import { mutations, SET_IS_CONNECTED } from './mutations';
 import { modules } from './modules';
+import { totalMileageReducer } from '../utils';
 
 export const USER = 'user';
-export const UNSYNCRONISED_DRIVES = 'unsyncedDrives';
-export const UNSYNCRONISED_DRIVES_TOTAL_MILEAGE = 'unsyncedDrivesTotalMileage';
+export const UNSYNCHRONISED_DRIVES = 'unsyncedDrives';
+export const UNSYNCHRONISED_DRIVES_TOTAL_MILEAGE = 'unsyncedDrivesTotalMileage';
 export const CARS = 'cars';
 export const LANGUAGE = 'language';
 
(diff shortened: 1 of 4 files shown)


frontend/src/store/index.js
```diff
@@ -9,6 +9,7 @@ import { modules } from './modules';
 
 export const USER = 'user';
 export const UNSYNCRONISED_DRIVES = 'unsyncedDrives';
+export const UNSYNCRONISED_DRIVES_TOTAL_MILEAGE = 'unsyncedDrivesTotalMileage';
```
adamgajzlerowicz: both line 11 and 12 are wrong.. It should be `UNSYNCHRONISED`

frontend/src/store/index.js
```diff
@@ -39,6 +40,10 @@ const store = new Vuex.Store({
   getters: {
     [IS_ONLINE]: state => state.isOnline,
     [UNSYNCRONISED_DRIVES]: state => state.unsyncedDrives,
+    [UNSYNCRONISED_DRIVES_TOTAL_MILEAGE]: state => state.unsyncedDrives.reduce(
```
adamgajzlerowicz: UNSYNCHRONISED please

adamgajzlerowicz: Why is the reduce needed twice? It is the same function so should be extracted.
Here getter is be polluting the global space- can't it be more local to the component?

frontend/src/views/DrivesView.vue
```diff
@@ -114,13 +126,17 @@ export default {
   methods: {
     ...mapActions(namespaces.drives, [apiActions.fetchDrives]),
     showDrive(id) {
-      this.driveVisible = id;
+      this.driveVisible = this.driveVisible === id ? null : id;
```
adamgajzlerowicz: I think there was a separate bug for that 👍

frontend/src/store/index.js
juliaanholcer: IMHO the purpose of the redux selectors is to derive the piece of state to the component and if some additional calculations are necessary they should be done by selector than in component.

frontend/src/store/index.js
juliaanholcer: Sorry for that.

frontend/src/views/DrivesView.vue
juliaanholcer: I'm not sure whether there is an issue related to that. I just wanted to make it more toggle-like than only openable.

adamgajzlerowicz:
Tested locally - looking good

PR #240, opened 2019-06-05
ManifestStaticFilesStorage connected with new static admin css cause …

magul:
…500 only when DEBUG=False

ManifestStaticFilesStorage would be also removed when Heroku will be cleanup but these two lines are hotfix for currently deployed production


magul:
CC: @Maciek246, @konradgalczynski07

PR #270, opened 2019-07-14
[263] Disable logging off when user is offline

### LINKED ISSUE #263: Users should not have option to log out when they are offline.
Hide the logout button when user is offline.

arturtamborski:
This PR will disable the link (specifically: disable <a> tag), but I'm happy to modify it to hiding completely if there's need for it (according to the PR https://github.com/CodeForPoznan/pah-fm/issues/263).

### DIFF
--- a/frontend/src/components/NavigationItems.vue
+++ b/frontend/src/components/NavigationItems.vue
@@ -26,6 +26,7 @@
       to="/logout"
       key="logout"
       class="username"
+      :class="{ offline: !isOnline }"
     >
       {{ $t('common.logout') }}
       <p>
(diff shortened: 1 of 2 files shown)


frontend/src/router/index.js
```diff
@@ -67,7 +67,7 @@ const openRoutes = [loginRoute.name, confirmationRoute.name];
 
 router.beforeEach((to, _from, next) => {
   // 404 if not route matches
-  if (to.name === pageNotFoundRoute) {
+  if (to.name === pageNotFoundRoute.name) {
```
adamgajzlerowicz: this is not right

frontend/src/components/NavigationItems.vue
```diff
@@ -90,4 +93,10 @@ export default {
   width: 100%;
   height: 100%;
 }
+
+.offline {
+  cursor: default;
+  pointer-events: none;
```
adamgajzlerowicz: awesome!

adamgajzlerowicz:
Excellent work with disabling the logout button.
I checked out your branch locally and I noticed that the redirect after the logout is not working correctly.
It would be awesome to close this part of flow and fix the logout feature too

Issue #251, opened 2019-07-08
Fuel consumption


monikakrawczak:
In Django Admin **exports**, add 'fuel consumption per drive'

Ctilhs:
@Ctilhs

magul:
Closed in #280

PR #280, opened 2019-07-17
[251] Added fuel_consumption field, to admin export.

Ctilhs:
https://github.com/CodeForPoznan/pah-fm/issues/251


magul:
@Ctilhs there's also merge conflict to be resolved. See: https://help.github.com/en/articles/resolving-a-merge-conflict-using-the-command-line

Ctilhs:
@magul In reference to our today's conversation. In model Drive fuel_consumption_per_drive is named this way to avoid using fuel_consumption name from from model Car.
Should we rename it to fuel_consumption in Drive model or keep it as fuel_consumption_per_drive?

Ctilhs:
@magul @arturtamborski 

During writing test for fuel concumption per drive I noticed an issue.
We don't have validation at model neither serializer layers.

Things directly related to fuel_consumption in Drive model which should be validated:
1. Is end_mileage bigger or equal to start_mileage
2. Are both start_mileage and end_mileage bigger or equal zero.

We might consider using PositiveIntegerField instead of IntegerField for fields start_mileage and end_mileage which would fix point 2.
In reqards to point 1 we should write cleaning/validating methods in appropriate model/serializer class.

backend/fleet_management/admin.py
```diff
@@ -18,22 +18,27 @@ def dehydrate_passengers(self, drive):
             for passanger in drive.passengers.all()
         )
 
+    fuel_consumption = Field(attribute='fuel_consumption')
```
magul: Do we need to make it explicit here?

backend/fleet_management/admin.py
```diff
@@ -18,22 +18,27 @@ def dehydrate_passengers(self, drive):
...
             "description",
             "start_location",
             "end_location",
+            "driver",
+            "fuel_consumption",
+            "car__plates",
             "project__title",
         )
 
+        export_order = fields
```
magul: Isn't it used by default?

backend/fleet_management/test_drives.py
```diff
@@ -185,3 +185,9 @@ def test_can_create_a_drive(self):
         )
         self.assertSetEqual({token.is_confirmed for token in tokens}, {False, False})
         self.assertSetEqual({token.is_ok for token in tokens}, {None, None})
+
+    def test_fuel_consumption_is_valid(self):
+        drive = self.drives[0]
```
magul: Multiple issues here.

First of all we could and should use factories here. Also if you don't want to use them, I don't really understand why you're usinng `self.drives[0]` and explicitly change it's `start_mileage` and `end_mileage`. That can potentialy make rest of test failing as they can depend on `mileage` values set in `setUp` method and nobody is assuring order of tests execution.

backend/fleet_management/admin.py
Ctilhs: Yes, by default, only fields from the database are imported. We want to additionally import @property of this model, for this we use import_export.fields.Field().

backend/fleet_management/admin.py
Ctilhs: Defining export_order is necessary to keep an earlier order. The change described here https://github.com/CodeForPoznan/pah-fm/pull/280#discussion_r308967346 makes fuel_consumption land as the first column which we don't want to.

backend/fleet_management/test_drives.py
Ctilhs: @magul As you suggested, I used factories, is it correct right now?
https://github.com/CodeForPoznan/pah-fm/pull/280/commits/f7e2a6e4b0ac5febc7ce711b9e29a7d4705c7909

PR #282, opened 2019-07-19
278 add is submitting

adamgajzlerowicz:


adamgajzlerowicz:
other issue with data submission was found

PR #293, opened 2019-08-06
Groups

adamgajzlerowicz:
- Added groups
- Assigned current users to Driver group
- Added route based permissions to routes (driver on all of them for now)


docker-compose.yml
```diff
@@ -34,11 +34,19 @@ services:
 
   db:
     image: postgres:9.5
+    ports:
+      - 5432:5432
```
arturtamborski: Can we change the host side port to something else? The reason behind it is that it might create a conflict for given developer who would want to start the project while also running some database on said port.
```suggestion
      - 6866:5432
```

docker-compose.yml
```diff
@@ -34,11 +34,19 @@ services:
...
     environment:
       POSTGRES_DB: pah-fm
       POSTGRES_USER: pah-fm
       POSTGRES_PASSWORD: pah-fm
 
+  adminer:
```
arturtamborski: What's the purpose of adding another container with (admittedly lightweight) image when it's not required for the project to operate?
Adminer's purpose is to act as an easy interface for database but this approach enforces it for every developer even if someone's using *any* other client, such as dbeaver, built-in pycharm DB tools or even raw PSQL shell.

If someone wants to interact with the database via adminer then she should spin it up with simple `docker run` command, no need to add it as a dependency.

backend/fleet_management/migrations/0011_add_default_groups.py
```diff
@@ -0,0 +1,32 @@
...
+from django.contrib.auth.models import Group
+from ..models import User
+
+
+from ..constants import Groups
+
+
+def create_groups(apps, schema_editor):
+    with connections['default'].cursor() as cursor:
+        cursor.execute(f"INSERT INTO auth_group (\"name\") values ('{Groups.Passenger.name}')")
```
arturtamborski: Could you use python instead of raw SQL insert? That's bit ugly in my opinion.
Even more, you can use Django's `RunSQL` instead of `cursor` but that would still be not needed.

You're using `Group` model few lines below anyway, so why not here?

```suggestion
        Group.objects.create(name=Groups.Passenger.name)
```

backend/fleet_management/permissions.py
```diff
@@ -0,0 +1,41 @@
+from django.contrib.auth.models import Group
+from rest_framework import permissions
+from .constants import Groups
+
+
+def is_in_group(user, group_name):
+    try:
+        return Group.objects.get(name=group_name).user_set.filter(id=user.id).exists()
```
arturtamborski: That's a potentially slow query which will be called very often due to the idea of groups acting as a permissions.
It's a reverse foreign key relation spanning across two tables. I'd like to at least see it cached somewhere or if that's not possible at the moment in reasonable way it might be good to rewrite it to something easier.

Here are some tips on that (maybe `select_related()`?):
https://medium.com/@ryleysill93/basic-performance-optimization-in-django-ebd19089a33f

Future idea: benchmarks!

docker-compose.yml
adamgajzlerowicz: Sounds good

docker-compose.yml
adamgajzlerowicz: I think it's helpful for development, but I agree it will be really easy to add it directly from terminal. Let me remove this

backend/fleet_management/migrations/0011_add_default_groups.py
adamgajzlerowicz: That sounds good :)

backend/fleet_management/permissions.py
adamgajzlerowicz: I am not sure - let's leave it for later :)

backend/fleet_management/migrations/0011_add_default_groups.py
adamgajzlerowicz: Also - thanks for the hint!

backend/fleet_management/permissions.py
arturtamborski: True, it's good enough considering current load.

backend/fleet_management/migrations/0011_add_default_groups.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db import migrations, connections
+from django.contrib.auth.models import Group
+from ..models import User
+
+
+from ..constants import Groups
+
+
+def create_groups(apps, schema_editor):
+    with connections['default'].cursor() as cursor:
```
arturtamborski: This line is not needed after my suggestion, can you remove it and indent the lines below to the left? Thanks!

adamgajzlerowicz: 👍

PR #298, opened 2019-08-08
Fix migrations

adamgajzlerowicz:


backend/docker-entrypoint.sh
```diff
@@ -3,6 +3,5 @@
 bash wait-for-it db:5432
 
 python3 manage.py migrate
-python3 manage.py create_admin hello@codeforpoznan.pl cfp123 --django-admin
-python3 manage.py create_admin ola@pah.org.pl pah123
+python3 manage.py populate_database
```
magul: I would prefer to have that as a separate command instead of part of starting backend container and would also prefer to leave `create_admin` as a separate script.

backend/fleet_management/factories.py
```diff
@@ -1,28 +1,28 @@
-import random
 import string
 from datetime import timedelta
 
-import factory
 import random
 from django.utils.timezone import now
-from factory import fuzzy
+from factory import fuzzy, DjangoModelFactory, Faker, LazyAttribute, lazy_attribute, \
```
magul: Looks like cosmetic change here. Don't see a reason for that, but I'm fine however it looks.

backend/fleet_management/factories.py
```diff
@@ -139,24 +139,26 @@ def drives(self, create, extracted, **kwargs):
...
-    car = factory.SubFactory(CarFactory)
+    driver = SubFactory(UserFactory)
+    project = SubFactory(ProjectFactory)
+    car = SubFactory(CarFactory)
     date = fuzzy.FuzzyDate((now() - timedelta(days=1000)).date())
     start_mileage = fuzzy.FuzzyInteger(1000000)
-    description = factory.Faker('text', max_nb_chars=1000)
+    description = Faker('text', max_nb_chars=1000)
     timestamp = fuzzy.FuzzyInteger(1, 999999999)
+    start_location = fuzzy.FuzzyChoice(CITIES)
```
magul: I would assume, that's the only reason of these changes.
We should use Faker here also. I would assume that it goes like:
```
start_location = Faker("city", locale="uk_UA")
```

backend/fleet_management/management/commands/populate_database.py
```diff
@@ -19,31 +19,49 @@ class Command(BaseCommand):
...
+        superuser.set_password('pah123')
+        superuser.country = 'UA'
+        superuser.save()
+
+        self.stdout.write(self.style.SUCCESS('Creating 5 cars'))
+        for _ in tqdm(range(5)):
             CarFactory.create()
 
-        self.stdout.write(self.style.SUCCESS('Creating 50 users'))
+        self.stdout.write(self.style.SUCCESS('Creating 5 users'))
```
magul: Do you have a reason to reduce the number of generated instances?

backend/fleet_management/management/commands/populate_database.py
```diff
@@ -19,31 +19,49 @@ class Command(BaseCommand):
     help = "Populates database with fake items."
 
     def handle(self, *args, **options):
-        """Populate database with fake objects."""
+        superuser = User()
```
magul: That's Python equivalence of:
```
var superuser = {};
superuser.is_active = true
superuser.is_superuser = true
superuser.is_staff = true
superuser.username = 'hello@codeforpoznan.pl'
superuser.email = 'hello@codeforpoznan.pl'
superuser.country = 'UA'
```
Would prefer to have that in init:
```
superuser = User(
	is_active=True,
	is_superuser=True,
	is_staff=True,
	username='hello@codeforpoznan.pl',
	email='hello@codeforpoznan.pl',
	country='UA',
)
```

backend/docker-entrypoint.sh
adamgajzlerowicz: If you think it will help - no problem. 
Can you please describe a scenario, where it would be required to run those two scripts separately?
Surely not on production  (password public). So this will be for development only.
And if it's only for development - than having two separate seed files makes little sense..

What do you think?

backend/fleet_management/management/commands/populate_database.py
adamgajzlerowicz: Yes - in development having 50 users is not required. 
Having these high numbers makes seeding longer. 
Also - the page loads longer on each refresh during development. 

High numbers in seeds are more of an edge case scenario for testers rather than good development numbers for developers.

backend/fleet_management/factories.py
adamgajzlerowicz: Looks good - let me update it

backend/fleet_management/management/commands/populate_database.py
adamgajzlerowicz: Lets discuss where this should go and update it accordingly then.

backend/fleet_management/factories.py
adamgajzlerowicz: `factory` was imported twice in two ways.
So eigher we go with `import factory` and we dont import `fuzzy` separetely or we do it as above.

backend/fleet_management/factories.py
adamgajzlerowicz: Updated - worked perfectly!

backend/docker-entrypoint.sh
magul: We discuss that offline with @arturtamborski and we agreed, that we want to have a separate script to create super users and another one to create objects from factories. Also as you limited number of instances created using factories when running `populate_database` it would make no sense to run that script whenever you start backend container.

backend/fleet_management/factories.py
magul: Don't recognize that as an error here, but I'm fine either way.

backend/fleet_management/management/commands/populate_database.py
magul: I just checked `create_admin` script and that's for me best approach how to create `User` objects:
```
user = User.objects.create_user(
    username=options['username'],
    email=options['username'],
    password=options['password'],
    is_superuser=options['django_admin'],
    is_staff=options['django_admin'],
    is_active=True,
)

```

backend/docker-entrypoint.sh
adamgajzlerowicz: Why do you want to have separate script for user creation?

How about actually dropping the entire schema on container start? This will make dev environment repeatable.

backend/fleet_management/management/commands/populate_database.py
adamgajzlerowicz: What is the difference?

backend/docker-entrypoint.sh
magul: Droping schema could be achived by `docker-compose rm db`. Personally believe that `make start` should not tear apart data available in previous runs.

backend/fleet_management/management/commands/populate_database.py
magul: I would say stylistic. I do not know anybody with extensive experience in Python that set attributes after creating an empty object as you did it here.

PR #299, opened 2019-08-08
remove traling slash in nginx configuration for non-Vue resources

adamgajzlerowicz:


magul:
Actually would also suggest to change that line in Ansilbe Nginx configuration: https://github.com/CodeForPoznan/pah-fm/blob/41d44a5a2578792e5480409dce07d86c90ded0bd/ansible/nginx.conf.j2#L10

adamgajzlerowicz:
Good idea!

PR #304, opened 2019-08-09
Add driver country column.

Ctilhs:
CountryField model by default returns country code.
I used dehydrate from import_export to access CountryField @property name which is full country name.

### DIFF
--- a/backend/fleet_management/admin.py
+++ b/backend/fleet_management/admin.py
@@ -18,6 +18,9 @@ class DriveResource(resources.ModelResource):
             for passanger in drive.passengers.all()
         )
 
+    def dehydrate_driver__country(self, drive):
+        return drive.driver.country.name
+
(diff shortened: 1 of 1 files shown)


backend/fleet_management/admin.py
```diff
@@ -18,6 +18,9 @@ def dehydrate_passengers(self, drive):
             for passanger in drive.passengers.all()
         )
 
+    def dehydrate_driver__country(self, drive):
+        return drive.driver.country.name
+
```
adamgajzlerowicz: I am wondering if it should be name or a symbol. We can change it later if needed anyways ;)

Ctilhs: That's right :) I assumed that since it is a report and it has not been clearly indicated that it is to be a country code, it is likely that it should be the full name of the country.

PR #306, opened 2019-08-10
Linter

adamgajzlerowicz:
pycodestyle is really limited. I added flake8 instead.
Now it will show errors when variables are defined but not used :)


backend/fleet_management/signals.py
```diff
@@ -1,39 +0,0 @@
-from django import dispatch
```
arturtamborski: What does this change (and the ones above) have in common with adding flake8?

adamgajzlerowicz: After linter was updated, it detected dead code. It wouldn't pass without a cleanup.

adamgajzlerowicz: @arturtamborski

backend/fleet_management/serializers.py
```diff
@@ -82,18 +81,6 @@ def create(self, validated_data):
             drive.passengers.set(passengers)
             drive.save()
 
-            for passenger in passengers:
```
arturtamborski: This code doesn't look like it's not being used. Also, flake8 did not report it as not needed when I run it on `develop`.
```
(venv) ~/self/pah-fm/backend(develop) $ flake8
./fleet_management/factories.py:6:1: F811 redefinition of unused 'random' from line 1
./fleet_management/signals.py:2:1: F401 'django.core.mail.EmailMessage' imported but unused
./fleet_management/signals.py:2:1: F401 'django.core.mail.get_connection' imported but unused
./fleet_management/signals.py:5:1: F401 'pah_fm.settings.EMAIL_ADDRESS' imported but unused
./fleet_management/signals.py:14:5: F841 local variable 'template' is assigned to but never used
./fleet_management/signals.py:15:5: F841 local variable 'subject' is assigned to but never used
./fleet_management/signals.py:17:5: F841 local variable 'drive' is assigned to but never used
./fleet_management/signals.py:18:5: F841 local variable 'driver' is assigned to but never used
./fleet_management/signals.py:19:5: F841 local variable 'token' is assigned to but never used
./fleet_management/signals.py:21:5: F841 local variable 'passenger' is assigned to but never used
```

backend/fleet_management/signals.py
arturtamborski: OK, but I still don't think that this code should be removed.

@magul / @monikakrawczak can you confirm the current situation with e-mails? Are we planning to send them out eventually? I see that we have a template mail and the code looks like it was used but was commented out due to production issues mentioned here: https://github.com/CodeForPoznan/pah-fm/pull/209 
Did anyone tried to take a look into it and fix it?

Anyway, I'd opt out for commenting this function out or adding this file to ignore list in flake8 instead of removing (what looks to be) a good, but unused code.

arturtamborski: Discussed this offline with @magul, seems like it is in fact not needed, so I'm giving a green light on this.

backend/fleet_management/serializers.py
arturtamborski: Discussed this offline with @magul, seems like it is in fact not needed, so I'm giving a green light on this.

Issue #262, opened 2019-07-10
"thank you" and confirmation screen


monikakrawczak:
Create "thank you" and confirmation screen or message after logging out for users to be sure that operation was successful.

arturtamborski:
@pniedzwiedzinski took care of this, closing.

PR #322, opened 2019-08-10
Add logout confirmation screen

pniedzwiedzinski:
I added logout confirmation screen (#262)


frontend/src/store/actions.js
```diff
@@ -48,7 +56,7 @@ export const actions = {
   [LOGOUT]({ commit }) {
     commit(mutations.SET_USER, null);
     deleteToken();
-    window.location.replace(login.path);
+    window.location.replace('/logout_success');
```
arturtamborski: does it have to be hardcoded?

frontend/src/views/SuccessfulLogoutView.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div>
+    <h1>Thank you!</h1>
```
adamgajzlerowicz: Please add to translations file and use translations (en/ua/pl)

frontend/src/store/actions.js
pniedzwiedzinski: The path could be imported from `routes`, but when I did that  Vuex store could not load for some reason 🤔I think it’s because `VueRouter` init is in the same file as routes - we could store them in different files.

pniedzwiedzinski: That didn't work either :/ I could do this it would be over-engineered. I will leave this as it is.

adamgajzlerowicz: I think that constants need to be refactored to separate file/files. This should  remove circular dependencies and fix this issue.
Let's do it in a separate pr ;)

adamgajzlerowicz:
Please fix linter

pniedzwiedzinski:
Should I update this?

https://github.com/CodeForPoznan/pah-fm/blob/4621f6130e6f1c971c143e8c9ea47afee76bac16/pah_behave/pah_behave/features/logout-testcase.feature#L1-L15

frontend/src/translations.json
```diff
@@ -63,6 +61,11 @@
       "username": "Ім'я користувача",
       "username_required": "Запит імені користувача"
     },
+    "logout": {
+      "thank_you": "Thank you!",
```
adamgajzlerowicz: Please use ukrainian - google translate is fine for now.

arturtamborski:
> Should I update this?
No, we do not require contributions with behave changes. Though unit tests would be cool but as far as I know we don't plan adding frontend testing suite.

pniedzwiedzinski:
Fixed

arturtamborski:
I really appreciate your work here though we still need few changes.

First thing - logout view is not aligned correctly. That's an issue because our app is used almost exclusively on mobile devices. Currently logout view is aligned to left whereas it should be on the middle.
image

Second thing - `/logout_success` is not blocked / hidden in any way. I can now open up the project website, _not_ log in and go straight to `/logout_success` and it will show me that I've logged out.
That's technically not an issue but it's also not desirable.

Feel free to ask for help and keep up the good work, we're almost there ;)

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
arturtamborski: thanks for adding this :)

pniedzwiedzinski:
There’s a Bug with menu on logout that I have not fixed yet

adamgajzlerowicz:
PR is looking good. Lets finish if there is anything left - we have been holding this pr for ages now. I am happy to fix whatever is left in a separate pr

pniedzwiedzinski:
I'm not sure about this last commit. It doesn't seem like the most elegant solution, but it works! What do you guys think? Is it ok?

arturtamborski:
Looks good enough for me though I'm not a front-end guy.

As for improving it I was thinking what if you'd display the sidebar only when user is logged in (in other words only when we have a session key).

You could remove the current functions `isLogin` and `isLogout` as they only check url and replace the `v-if` to something like this: 
`v-if=isLoggedIn`.

This function could then check if said token was in localStorage then user could be considered as logged in so we could show the sidebar.

pniedzwiedzinski:
It should be ready for merge.

Issue #256, opened 2019-07-08
Passenger - country


monikakrawczak:
Passengers may be assigned to one country.
If the passenger is not assigned to the country, the passenger should be visible in every country.

monikakrawczak:
Passenger list for driver only in the same country (or not set)

Ctilhs:
@Ctilhs

Ctilhs:
https://github.com/CodeForPoznan/pah-fm/pull/326

PR #326, opened 2019-08-13
#256 Add country to Passenger

Ctilhs:
Some test issues need to be discussed, then tests will be added.
@arturtamborski


backend/fleet_management/models.py
```diff
@@ -41,6 +41,7 @@ class Passenger(models.Model):
     first_name = models.CharField(max_length=60, blank=False)
     last_name = models.CharField(max_length=60, blank=False)
     email = models.EmailField(blank=False)
+    country = CountryField(blank_label='(select country)', null=False, default='')
```
adamgajzlerowicz: How about using NULL for value that is not set?
Setting it to an empty string means that it was set to this value on purpose,

Ctilhs: Fixed https://github.com/CodeForPoznan/pah-fm/pull/326/commits/bdca70d9425f2b60d40adbf7a4ed251d2363c1a7

PR #336, opened 2019-08-23
add actual rebuild

adamgajzlerowicz:


Makefile
```diff
@@ -18,7 +18,12 @@ remove:  ## Stop and remove backend & frontend containers
 	make remove-backend
 	make remove-frontend
 
-rebuild:  ## Rebuid backend & frontend containers
+rebuild:  ## Rebuid application
+    docker-compose down
```
magul: I would suspect, that you're mixing spaces and tabs here. Makefile requires tabs.

adamgajzlerowicz: is there a linter for make files? xD

PR #337, opened 2019-08-23
added adminer

adamgajzlerowicz:


Makefile
```diff
@@ -14,6 +14,9 @@ build:  ## Build backend & frontend containers
 	make build-backend
 	make build-frontend
 
+adminer: ## run database adminer
+	docker run -d -p=8082:8080 --network="pah" adminer
```
m-recz: How about:
docker run -d -p 8082:8080 --network "pah" adminer

adamgajzlerowicz: Thanks - I updated the pr

PR #342, opened 2019-08-27
308 migrate passengers to groups - step one

adamgajzlerowicz:


backend/fleet_management/migrations/0016_auto_20190829_2155.py
```diff
@@ -0,0 +1,53 @@
...
+            password=password,
+            is_superuser=False,
+            is_staff=False,
+            country=passenger.country,
+            is_active=True,
+            first_name=passenger.first_name,
+            last_name=passenger.last_name
+        )
+
+        user.save()
```
arturtamborski: That's not needed, see https://docs.djangoproject.com/en/2.2/ref/contrib/auth/#django.contrib.auth.models.UserManager.create_user

backend/fleet_management/migrations/0016_auto_20190829_2155.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_users_from_passengers(apps, schema_editor):
+    passenger_group = Group.objects.get(name=Groups.Passenger.name)
+
+    passengers = Passenger.objects.all()
+
+    for index, passenger in enumerate(passengers):
+        fake = Faker()
+        password = fake.password()
```
arturtamborski: why are you faking passwords? Wouldn't this effect in loosing user account passwords?

adamgajzlerowicz: We are changing passengers (no password) to full users, assigned to the Passenger group

Because we are removing a Passenger model we can't do this change in one step :(
Thus I am breaking this down to two PRs. I am having some problems with the second one though... I will try to sort it out before Wednesday though

Issue #40, opened 2018-11-07
Drives: sync once network is available


jacol12345:
Once the user is logged and on-line the routes they created while offline are synced with the backend.

adamgajzlerowicz:
Blocked by https://github.com/CodeForPoznan/pah-fm/issues/106 / https://github.com/CodeForPoznan/pah-fm/pull/110

monikakrawczak:
I splited 2 new issues from that one (#148, #149).
This issue should be limited only to sync.

adamgajzlerowicz:
already merged
https://github.com/CodeForPoznan/pah-fm/pull/116

PR #347, opened 2019-09-24
Tests refactoring

Ctilhs:
No new tests were added, only refactored existing ones.


backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,95 @@
...
+            "project": {"id": self.project.id},
+        }
+
+    def test_401_for_unlogged_user(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_can_retrieve_only_my_drives(self):
+        new_driver = UserFactory(
+            username="JessicaDownson",
```
arturtamborski: Do you have to explicitly specify this information? I don't see it being used below, so thought we could let it be generated by `UserFactory` instead.

Ctilhs: Agree, it's not required to specify them.
Fixed it and removed response.json which I was using for debugging.
https://github.com/CodeForPoznan/pah-fm/pull/347/commits/cfcd74789b1f42ee408495d5e4ea2b49ee903246

arturtamborski: Good, thanks, but I still see one file `backend/serializer.json` in PR which looks like is also not needed.

backend/requirements.txt
```diff
@@ -0,0 +1,56 @@
+asn1crypto==0.24.0
```
arturtamborski: why is this file green? It doesn't look different than current one in `master`

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_401_for_unlogged_user(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_can_retrieve_only_my_drives(self):
+        new_driver = UserFactory()
+
+        new_driver.groups.set(Group.objects.filter(name=Groups.Driver.name))
+
```
arturtamborski: nitpick: we don't need these newlines ;)

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,43 @@
...
+        ]
+
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_get_all_cars(self):
+        self.client.force_login(self.user)
+        res = self.client.get(self.url)
+        cars = res.data
```
arturtamborski: You're using `data.json()` in `#40`th line, is there any reason that you've used `res.data` here instead?

Ctilhs: Actually no, it behaves the same just missed that.

backend/fleet_management/tests/test_drives.py
Ctilhs: Ah true, same like with response.json, removed.

backend/fleet_management/tests/test_drives.py
Ctilhs: Fixed :+1:

arturtamborski:
Ok, so for some reason magul's comments have not shown up here

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertEqual(drives[0]["id"], new_drive[0].id)
+
+    @override_settings(EMAIL_BACKEND="django.core.mail.backends.filebased.EmailBackend")
+    def test_can_create_a_drive(self):
+        """This test allows to verify email template."""
+        self.client.force_login(self.driver)
+        res = self.client.post(self.url, data=self.payload, format="json")
+        drive = Drive.objects.get(pk=res.data["id"])
+        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(DriveSerializer(drive).data, res.data)
```
arturtamborski: This comment will be unfair from my side, because I proposed that solution in the first place but can you please change it so that you don't compare the `res.data`/`res.json()` to the output of `DriveSerializer`?
The current assertion is kinda pointless, you're creating some models with factories which is fine, then you query that data back via `get()` request and that's also fine but the returned value from given endpoint is filtered trough `Serializer` and we're comparing the response to that same serializer, so whatever comes out of it will be valid as far as this assertion thinks.
So yeah, sorry for that, I must have not thought this trough this idea. Can you change it so that you're asserting that  `res.json()` is equal to some hardcoded dict instead? I'll be happy to assist as well.

backend/requirements.txt
arturtamborski: This file is not needed. If you want to add some requirements please do so in `backend/requirements/{base,dev}.txt`.

magul:
I'm still writting them @arturtamborski :/

backend/fleet_management/factories.py
```diff
@@ -148,7 +148,7 @@ class Meta:
     driver = SubFactory(UserFactory)
     project = SubFactory(ProjectFactory)
     car = SubFactory(CarFactory)
-    date = fuzzy.FuzzyDate((now() - timedelta(days=1000)).date())
+    date = fuzzy.FuzzyDate((now() - timedelta(days=1000)).date(), now().date())
```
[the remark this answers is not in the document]
Ctilhs: That's right it's finally the same. Fixed.

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,43 @@
...
+
+from fleet_management.constants import Groups
+from fleet_management.factories import CarFactory, UserFactory
+from fleet_management.serializers import CarSerializer
+
+
+class CarsApiTest(APITestCase):
+    def setUp(self):
+        self.url = reverse("cars")
+        self.user = UserFactory(username="Admin")
```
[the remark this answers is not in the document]
Ctilhs: Agree, fixed.

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,43 @@
...
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_get_all_cars(self):
+        self.client.force_login(self.user)
+        res = self.client.get(self.url)
+        cars = res.data
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(cars), 3)
```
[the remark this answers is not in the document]
Ctilhs: Fait point, fixed.

backend/requirements.txt
Ctilhs: Removed.

backend/fleet_management/tests/test_drives.py
Ctilhs: @arturtamborski I changed it, please let me know if it's all correct now.

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+            size=10,
+            car=self.car,
+            driver=self.driver,
+            date=date.today().isoformat(),
+            project=self.project,
+        )
+
+        self.drives[0].passengers.set(self.passengers)
+
+        self.payload = {
```
magul: It looks like you use that symbol only once in `test_can_create_a_drive`. Until it will be used multiple time in different test cases I would suggest storing that in `test_can_create_a_drive` as that make this test case a lot more readable (I don't need to figure out what `self.payload` is being in the context of `test_can_create_a_drive`).

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+        self.client.force_login(new_driver)
+        res = self.client.get(self.url)
+        drives = res.data
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(drives), 1)
+        self.assertEqual(drives[0]["id"], new_drive[0].id)
+
+    @override_settings(EMAIL_BACKEND="django.core.mail.backends.filebased.EmailBackend")
+    def test_can_create_a_drive(self):
+        """This test allows to verify email template."""
```
magul: AFAIR we removed email sending functionality altogether so this description is wrong.

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+        ]
+
+        self.client.force_login(new_driver)
+        res = self.client.get(self.url)
+        drives = res.data
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(drives), 1)
+        self.assertEqual(drives[0]["id"], new_drive[0].id)
+
+    @override_settings(EMAIL_BACKEND="django.core.mail.backends.filebased.EmailBackend")
```
magul: We don't send emails at all, so probably we could remove that `override_setting` decorator?

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_can_create_a_drive(self):
+        """This test allows to verify email template."""
+        self.client.force_login(self.driver)
+        res = self.client.post(self.url, data=self.payload, format="json")
+        drive = Drive.objects.get(pk=res.data["id"])
+        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
+        self.assertSetEqual(
```
magul: General note here (as that is how that test looked like in past and let say, I'm fine with it) - we have 4 places, that contains some information about drive to be created:
* `self.drive` which is a place, you save drive to be used in the payload
* `self.payload` that actually contains the serialized drive that would be sent in the request
* `res.data` which contains what Django responded with
* `Drive.objects.get(pk=res.data["id"])` which contains Django internal model object.

We should contemplate that and make a general decision, what in fact we want to compare and what value does it provide for us?

magul: (also that could be viewed as discussion what does it mean to test serializer and what does it mean to test view, we will move that discussion offline on next hacknight)

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_can_retrieve_only_my_drives(self):
+        new_driver = UserFactory()
+        new_driver.groups.set(Group.objects.filter(name=Groups.Driver.name))
+        new_drive = [
```
magul: The main benefit of using factories is that you don't need to provide all values to it. So I would assume that you can simplify that line to:
```
DriveFactory(driver=new_driver)
```
and rest should be created automagically for you.

Also you consume only `id` of that `Drive` and there's only one drive, so I would assume that can be done like
```
my_drive_id = DriveFactory(driver=new_driver).id
```
and then in last line
```
self.assertEqual(drives[0]["id"], my_drive_id)
```

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+    DriveFactory,
+    PassengerFactory,
+    CarFactory,
+    ProjectFactory,
+    UserFactory,
+)
+from fleet_management.models import Drive
+
+
+class DrivesApiTest(APITestCase):
```
magul: You abandoned `test_get_all_drives`. Is that on purpose?

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.drives = DriveFactory.create_batch(
+            size=10,
+            car=self.car,
+            driver=self.driver,
+            date=date.today().isoformat(),
+            project=self.project,
+        )
+
+        self.drives[0].passengers.set(self.passengers)
```
magul: Having that in factories: https://github.com/CodeForPoznan/pah-fm/blob/250d7d92763bbac059efd62e0454822979fc1a4d/backend/fleet_management/factories.py#L163 I would assume that we can do that like:
```
DriveFactory(... passengers=self.passengers)
```

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+        self.project = ProjectFactory(country="UA")
+        self.driver = UserFactory()
+        self.driver.groups.set(Group.objects.filter(name=Groups.Driver.name))
+
+        self.drives = DriveFactory.create_batch(
+            size=10,
+            car=self.car,
+            driver=self.driver,
+            date=date.today().isoformat(),
+            project=self.project,
```
magul: Do we rely on `self.project` in any way? (and `self.car`, and `date.today().isoformat()`)

I guess not, so we could remove `self.car` and `self.project` altogether and build that as
```
self.drives = DriveFactory.create_batch(size=10, driver=self.driver)
```

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class DrivesApiTest(APITestCase):
+    def setUp(self):
+        self.url = reverse("drives")
+        self.passengers = PassengerFactory.create_batch(size=10)
+        self.car = CarFactory()
+        self.project = ProjectFactory(country="UA")
+        self.driver = UserFactory()
+        self.driver.groups.set(Group.objects.filter(name=Groups.Driver.name))
```
magul: You use that at least twice in that `TestCase`, so maybe `self.driver_group = Group.objects.filter(name=Groups.Driver.name))`?

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,42 @@
...
+from fleet_management.factories import CarFactory, UserFactory
+from fleet_management.serializers import CarSerializer
+
+
+class CarsApiTest(APITestCase):
+    def setUp(self):
+        self.url = reverse("cars")
+        self.user = UserFactory()
+        self.user.groups.set(Group.objects.filter(name=Groups.Driver.name))
+        self.cars = [
```
magul: Why not
```
self.cars = CarFactory.create_batch(3, country=self.user.country)
```
and then in `test_search_by_plate`
```
url_params = urlencode({"search": self.cars[0].plate})
```
?

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_get_all_cars(self):
+        self.client.force_login(self.user)
+        res = self.client.get(self.url)
+        cars = res.data
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(CarSerializer(self.cars, many=True).data, cars)
```
magul: We had a discussion with @arturtamborski about using serializers to prepare expected data to be asserted. I believe we should omit using it like that. (that's also a part of the discussion, what does it mean to test serializer, and what it takes to test view)

backend/fleet_management/tests/test_drives.py
Ctilhs: @magul 
Yes, get_queryset in DriveView filter Drives in response by user_id. 
From what I understand if there is more than one driver and these drivers have any drives then it's not possible to get all drives by them.

So I conclude only test_can_retrieve_only_my_drives is testing what actually is happening in DriveView.

backend/fleet_management/tests/test_drives.py
Ctilhs: Agree, moved it to `test_can_create_a_drive.`

backend/fleet_management/tests/test_drives.py
Ctilhs: Fixed.

backend/fleet_management/tests/test_drives.py
Ctilhs: Fixed.

backend/fleet_management/tests/test_drives.py
Ctilhs: Agree, fixed.

backend/fleet_management/tests/test_drives.py
Ctilhs: Thanks! I didn't know it's working that way. Fixed.

backend/fleet_management/tests/test_drives.py
Ctilhs: I have checked, it's working without it. Fixed.

backend/fleet_management/tests/test_drives.py
Ctilhs: Good point, updated in `setUp` and `test_can_retrieve_only_my_drives`.

backend/fleet_management/api.py
```diff
@@ -56,7 +56,7 @@ class DriveView(generics.ListCreateAPIView):
     required_groups = all_driver_methods
     serializer_class = DriveSerializer
     filter_backends = (filters.OrderingFilter,)
-    ordering = ('-date',)
+    ordering = ('-date', 'id',)
```
magul: Ordering of `-date` means that drives will be ordered by date descending, if we want to obey that rule we should use `-id` here.

backend/fleet_management/tests/test_drives.py
magul: It's used in this method only, so no reason to set it as class instance attribute (so no `self`, just `payload =`)

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,87 @@
...
+        new_driver.groups.set(self.driver_group)
+        my_drive_id = DriveFactory(driver=new_driver).id
+
+        self.client.force_login(new_driver)
+        res = self.client.get(self.url)
+        drives = res.data
+
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(drives), 1)
+        self.assertEqual(drives[0]["id"], my_drive_id)
```
magul: I've just re-reviewed  `test_can_retrieve_only_my_drives` and `test_get_all_drives` and it occurs, that `test_can_retrieve_only_my_drives` was using `self.driver` to login to the application and check that others drivers' drives are excluded, so I would suppose that we will have have something like
```
self.client.force_login(self.driver)
...
self.assertEqual(len(drives), len(self.drives))
for drive in drives:
    self.assertEqual(drive["driver"]["id'], self.driver.id) 
```

backend/fleet_management/tests/test_cars.py
Ctilhs: Totaly right, fixed.

backend/fleet_management/api.py
Ctilhs: It's a misconception, it's not necessary to change this.

backend/fleet_management/tests/test_drives.py
Ctilhs: I think I fixed it in the next commit.

arturtamborski:
@magul please do a review one last time for this because from what I saw all your comments were addressed pretty carefully. I think that this PR is ready to be merged (hence three approvals :smiley:)

backend/fleet_management/tests/test_drives.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(drive.end_location, res.data["end_location"])
+
+    def test_fuel_consumption_is_valid(self):
+        drive = DriveFactory(start_mileage=100300, end_mileage=100500)
+        drive.car.fuel_consumption = 9.73
+        self.assertEqual(drive.fuel_consumption, 19.46)
+
+    def test_diff_mileage(self):
+        drive = DriveFactory(start_mileage=100300, end_mileage=100800)
+        self.assertEqual(drive.diff_mileage, 500)
```
magul: newline

backend/fleet_management/tests/test_cars.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_get_all_cars(self):
+        self.client.force_login(self.user)
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(self.cars), len(res.data))
+        self.assertEqual({c["plates"] for c in res.data}, {c.plates for c in self.cars})
```
magul: Is there a reason abandoning more extensive comparison here?

magul: I remember we had that conversation few weeks ago. I will try to approach this issue with maybe another solution.

PR #348, opened 2019-10-01
Feature/passenger field

### LINKED ISSUE #233: New route - passenger field
Typling to the box should filter selector - now it doesn't provide any value.

pniedzwiedzinski:
I added filtering on passenger field in the new drive form (using `v-select`)

This PR closes #233


pniedzwiedzinski:
Does anyone know why there are old commits showing up?

frontend/src/main.js
```diff
@@ -1,11 +1,12 @@
 import 'bootstrap-vue/dist/bootstrap-vue.css';
 import 'bootstrap/dist/css/bootstrap.css';
-import FlagIcon from 'vue-flag-icon';
```
magul: Is there's some internal logic of changes in that file? (as it looks rather as import sorting and some formating)

frontend/src/views/DriveFormView.vue
```diff
@@ -312,4 +321,28 @@ export default {
 .wrapper {
   @include m(2);
 }
+
+.select {
```
magul: Do we need so extensive styling?

frontend/src/main.js
pniedzwiedzinski: Should I revert that?

magul: No, I'm just curious why such changes.

pniedzwiedzinski: I was changing there but I’ve found a simpler solution ;)

frontend/src/views/DriveFormView.vue
pniedzwiedzinski: No ;) fixed

arturtamborski:
Tested locally right after latest commit (fixes local storage caught by @magul) and it works fine.
Green light from my side.

PR #349, opened 2019-10-02
Create first Behave feature

kruczyna:
Behave kickoff - created framework and first feature files to test login page

### DIFF
--- a/test_automation/features/steps/login.py
+++ b/test_automation/features/steps/login.py
@@ -1,0 +1,49 @@
+import features.steps.storage as storage
+import features.steps.login_selectors as login_selectors
+
+from features.steps.common_actions import short_wait
+
+from behave import given, step, then, when
+from behave.runner import Context
+
+
+@given("I have url")
+def get_form_url(context: Context) -> None:
+    context.browser.get(storage.form_url)
+
+
+@given("I am on login page")
+def login_page(context: Context) -> None:
+    context.browser.find_element(*login_selectors.main_logo)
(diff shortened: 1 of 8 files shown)


test_automation/features/steps/login.py
```diff
@@ -0,0 +1,49 @@
...
+def input_password(context: Context, password: str) -> None:
+    password_field = context.browser.find_element(*login_selectors.password_field)
+    password_field.click()
+    password_field.send_keys(password)
+
+
+@when("I click login button")
+def click_login(context: Context) -> None:
+    login_button_enabled = context.browser.find_element(*login_selectors.login_button_enabled)
+    login_button_enabled.click()
```
arturtamborski: `input_username()`, `input_password()` and `click_login()` look very much the same to me - that's a good opportunity to take the repeated code out to another helper function, somewhere in `common_actions`.
Here's a rough draft of that:
```
# common_actions.py
def click_on(context: Context, element: Tuple[str, str]) -> WebElement:
    return context.browser.find_element(*element).click()
```
```
# login.py
@step('I input "{password}" password')
def input_password(context: Context, password: str) -> None:
    click_on(*login_selectors.password_field).send_keys(password)
```

It looks like the `find_element()/click()` combo is the default for pretty much every step it would be good to have it shortened to side-function helper, just like the `short_wait()` function.

arturtamborski:
I see that you've created another PR with similar changes at #358, and this one was not worked one for some time now, so I'm closing it.

PR #353, opened 2019-10-10
Create dev users

### LINKED ISSUE #352: Populate database
I was thinking about the populate database function that we have. I've come to conclusion that when generating all data we should also create some constant test users. Before that I was using the admin user (hello@codeforpoznan.pl) for testing how the application works, but the truth is that admin account differs from driver and from passenger. I was checking how implement it and I have a solution. We could add in `backend/fleet_management/management/command/populate_database.py` 2 fixed users - 1 driver and 1 passenger. What do you guys think about it?

arturtamborski:
This pull request renames command `create_admin` to `create_user` and also adds two new permanent users for testing purposes.
Closes #352


arturtamborski:
Please revisit this @magul, I still think that it would be good idea to have some _constant_ test accounts for driver and passenger, because the factory ones (although still very good and real) are quite random which slows down any manual test of the website from user's perspective.

The default users are no good for this, because they don't have first/last name and they don't belong to any group. They are by definition admin users so they don't take part in the flow of application.

magul:
We can have that conversation next Wednesday but even if we want to have driver/passenger with constant email address I would prefer to have them created in `make populate-database` (we could use factories to that).

What do you think about that? What do you think about that @kruczyna ?

PR #354, opened 2019-10-11
[311] add rsa field to user model

Ctilhs:


Ctilhs:
I'm working on it to fix it.

arturtamborski:
There's too much trash in here, we'll recreate it the old way :)

PR #360, opened 2019-10-23
Add rsa fields to User model.

Ctilhs:


arturtamborski:
This change may fail locally, because migration no. 11 is kinda broken, but we have a proper fix in place in another Pull Request #350. Everything else should be good.

PR #361, opened 2019-10-23
Feature/passenger view

### LINKED ISSUE #317: New Passenger form
Based on current user role (== passenger) show the new form two fields:
* input field for drive hash (input validation lengh == 6)
* read only field for sign value

### LINKED ISSUE #355: List of errors actually an object
In new drive form, function `validateForm` creates list of error by joining objects. Wouldn't it be simpler if we would have an array of errors?

https://github.com/CodeForPoznan/pah-fm/blob/develop/frontend/src/views/DriveFormView.vue#L259-L284

### LINKED ISSUE #381: Value written in start mileage / end mileage are not removable
image
There's something wrong with these two number fields - I can't remove value from them with delete or backspace. I can't remove digits, I can only add new ones. The only way of decreasing the value is to use the "arrow down" button, on the right side of the input box.

pniedzwiedzinski:
This PR adds view for passengers.

This PR will close #317 and fix #355 and fix #381


pniedzwiedzinski:
It should be okay now.

arturtamborski:
Sidebar is empty, I can't change language or log in/out of the app. Can you please check it again?

pniedzwiedzinski:
Are you sure that you are logged as driver or passenger?

arturtamborski:
Ok, I see that you've fixed the issue with sidebar which is good, but I still noticed few problems with this PR:
1. When I'm logged in as a Passenger I still can submit a drive (the /drive form). I'm also redirected to it right after the log in.
2. Input form in /passenger allows incorrect text, for example letters. You can fix it by changing input type to `number`.
3. The confirmation code that was calculated is a bit too small and there's no reason for it to be like it. 
4. I can go straight to /drives even though I'm logged in as Passenger.

pniedzwiedzinski:
> 3. The confirmation code that was calculated is a bit too small and there's no reason for it to be like it.

What do you mean by that?

arturtamborski:
I meant the size of the numbers

pniedzwiedzinski:
On last meeting didnt’t we discuss that hashes should be length of 6?

Patryk Niedźwiedziński

arturtamborski:
> > 1. The confirmation code that was calculated is a bit too small and there's no reason for it to be like it.
> 
> What do you mean by that?

I meant the size of numbers, I think that they should be bigger thus more readable (especially on mobiles).

arturtamborski:
> On last meeting didnt’t we discuss that hashes should be length of 6? Patryk Niedźwiedziński
> […](#)
> On Nov 27, 2019, at 9:47 AM, Artur Tamborski ***@***.***> wrote: I meant the size of the numbers — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Yes, we did and you're right, the length should be 6. I'm just suggesting that the font is a bit too small in my opinion.

pniedzwiedzinski:
Ah, dang, yes, my fault

pniedzwiedzinski:
image
Something like this would be okay?

frontend/src/components/ErrorList.vue
```diff
@@ -0,0 +1,29 @@
...
+    errors: {
+      type: Array,
+      required: true,
+    },
+  },
+};
+</script>
+
+<style scoped>
+.error::first-letter {
```
Bosmanfrx:
```suggestion
.error:first-letter {
```

frontend/src/components/MainForm.vue
```diff
@@ -0,0 +1,49 @@
...
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import '../scss/base';
+
+.wrapper {
+  @include m(2);
```
Bosmanfrx: Wouldn't it better to just add class m2? It will result in duplicated css across project

frontend/src/components/NavigationItems.vue
```diff
@@ -1,96 +1,33 @@
...
+      }}
     </b-nav-item>
-  </b-nav>
+  </div>
 </template>
 
 <script>
-import { mapActions, mapGetters, mapState } from 'vuex';
-import * as actions from '../store/actions';
+import flatMap from 'array.prototype.flatmap'
```
Bosmanfrx: As we updated to core-js v3 - is another polyfill needed? If it is, everything is ok

frontend/src/mixins/FormMixin.js
```diff
@@ -0,0 +1,69 @@
...
+    validateForm(validator = undefined) {
+      /**
+       * In validator argument you should pass a function that takes `this.form` as argument and
+       * returns an array of errors (strings). In that function you also have access to
+       * `this.isInvalid` object, in which you can control highlighting of field.
+       */
+      this.isInvalid = {};
+      this.listOfErrors = [];
+
+      this.listOfErrors = this.requiredFields.reduce((acc, field) => {
```
Bosmanfrx: We should not use reduce to produce another array. We probably want map and filter

frontend/src/mixins/GroupGuardMixin.js
```diff
@@ -0,0 +1,29 @@
+import store from '../store';
+import { groupBasedRoutes } from '../router';
+import { reverseObject } from '../services/reverseObject';
+
+const { user } = store.state;
+const groups = user ? user.groups.map(({ name }) => name.toLowerCase()) : [];
```
Bosmanfrx: nice!

frontend/src/router/index.js
```diff
@@ -92,6 +141,23 @@ router.beforeEach((to, _from, next) => {
   if (userLoggedIn && to.name === loginRoute.name) {
     return next({ path: homeRoute.path });
   }
+
+  // Guard routes based on groups
+  if (userLoggedIn && allGroupBasedRoutes.includes(to.name)) {
+    const availableRoutes = store.state.user.groups
+      .reduce(
```
Bosmanfrx: ditto

frontend/src/services/reverseObject.js
```diff
@@ -0,0 +1,26 @@
...
+ * }
+ * //will be converted to:
+ * {
+ *  value1: key1,
+ *  value2: key2,
+ *  value3: key2
+ * }
+ */
+export const reverseObject = objectToReverse =>
+  Object.keys(objectToReverse).reduce(
```
Bosmanfrx: It would be easier with `Object.entries`, since you have kay and value there

frontend/src/views/PassengerSubmitView.vue
```diff
@@ -0,0 +1,52 @@
...
+    return next({ path: '/passenger' });
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import '../scss/base';
+
+.wrapper {
+  @include m(2);
```
Bosmanfrx: ditto

frontend/src/components/NavigationItems.vue
pniedzwiedzinski: I couldn't get core-js to work 😞. So that's why I'm importing this

arturtamborski:
Great and consistent work on that, good job :) Please resolve the conflicts above so it could be ready for a merge.

frontend/src/components/Sidebar.vue
```diff
@@ -0,0 +1,88 @@
...
+  margin: 60px auto;
+  width: 80px;
+}
+
+.out-link-image {
+  width: 100%;
+  height: 100%;
+}
+
+.offline {
```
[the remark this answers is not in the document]
pniedzwiedzinski: Here:
https://github.com/pniedzwiedzinski/pah-fm/blob/6a502d8950fac1b25bdba1b9e66ac94f8582edd9/frontend/src/components/Sidebar.vue#L28

pniedzwiedzinski:
Ready for merge

PR #368, opened 2019-11-19
[335] Create key generation for RSA-based PKI

### LINKED ISSUE #335: Create key generation for RSA-based PKI
Also trigger that generation when creating new user and store RSA keys in `User` model.

# Basic requirement
* message length 6 decimal characters (which mean `10^6 ~= 2^20`, so  `p*q = n` 20 bits long).

# Implementation
Proof of concept implementation (need to have that only in Python, JS will only require `sign` and `verify` functions):
```
import math
import random


GLOBAL_BITSIZE = 24
GLOBAL_UPPER_BOUNDARY = 2 ** GLOBAL_BITSIZE - 1
EXPONENT = 257


def get_random_odd(bitsize):
    return random.randint(0, 2 ** bitsize - 1) | 1


def is_prime(number):
    upper_bondary = math.floor(math.sqrt(number)) + 1
    for i in range(2, upper_bondary + 1):
        if number % i == 0:
            return False
    return True


def get_prime(bitsize):
    while True:
        integer = get_random_odd(bitsize)

        if is_prime(integer):
            return integer


def is_acceptable_p_q(p, q):
    if p == q:
        return False
    return p * q >= (GLOBAL_UPPER_BOUNDARY + 1) // 2


def get_p_q():
    shift = 3
    pbits = GLOBAL_BITSIZE // 2 + shift
    qbits = GLOBAL_BITSIZE // 2 - shift

    p = get_prime(pbits)
    q = get_prime(qbits)

    change_p = False
    while not is_acceptable_p_q(p, q):
        if change_p:
            p = get_prime(pbits)
        else:
            q = get_prime(qbits)
        change_p = not change_p

    return max(p, q), min(p, q)


def inverse(a, b):
    x = 0
    y = 1
    lx = 1
    ly = 0
    oa = a
    ob = b
    while b != 0:
        q = a // b
        (a, b) = (b, a % b)
        (x, lx) = ((lx - (q * x)), x)
        (y, ly) = ((ly - (q * y)), y)
    if lx < 0:
        lx += ob
    if ly < 0:
        ly += oa
    return lx


def calculate_keys_custom_exponent(p, q):
    phi_n = (p - 1) * (q - 1)
    d = inverse(EXPONENT, phi_n)

    if (EXPONENT * d) % phi_n != 1:
        raise ValueError

    return EXPONENT, d


def get_keys():
    while True:
        p, q = get_p_q()
        try:
            e, d = calculate_keys_custom_exponent(p, q)
            break
        except ValueError:
            pass

    n = p * q
    return {
        "pub": {"n": n, "e": e},
        "priv": {"n": n, "e": e, "d": d, "p": p, "q": q}
    }


def check_sign(message, sign, key):
    return pow(sign, key["e"], key["n"]) == message % key["n"]


def sign(message, key):
    return pow(message, key["d"], key["n"])
```

arturtamborski:
Closes #335.


backend/fleet_management/crypto.py
```diff
@@ -0,0 +1,104 @@
...
+    for i in range(2, floor(sqrt(number)) + 2):
+        if number % i == 0:
+            return False
+
+    return True
+
+
+def find_prime(bits: int) -> int:
+    """Returns prime number with specified amount of bits."""
+    prime = randbits(bits) | 1
```
magul: That probably can be written in a lot compact way:

```python
def find_prime(bits: int) -> int:
    """Returns prime number with specified amount of bits."""
    while True:
        prime = randbits(bits) | 1
        while prime.bit_length() <= bits:
            if is_prime(prime):
                return prime
            prime += 2
```

backend/fleet_management/crypto.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def find_p_q_phi() -> (int, int, int):
+    """Returns RSA key components."""
+    limit = (2 ** settings.RSA_NUMBER_OF_BITS) // 2
+    p_bits = settings.RSA_NUMBER_OF_BITS // 2 + 3
+    q_bits = settings.RSA_NUMBER_OF_BITS // 2 - 3
+    p, q = find_prime(p_bits), find_prime(q_bits)
+
+    other = False
```
magul: I would suggest to use something more meaningfull here. Maybe `other_p` to indicate that we're going to re-create `p`.

backend/fleet_management/crypto.py
```diff
@@ -0,0 +1,104 @@
...
+        y0, y1 = y1, y0 - q * y1
+
+    if x0 < 0:
+        x0 += ob
+
+    if y0 < 0:
+        y0 += oa
+
+    # when a and b are not relatively prime
+    if a != 1 or (oa * x0) % ob != 1:
```
magul: That's rather `c` (or `golang`) style. Python should raise exception here.

magul: (or is that proper result based on module algebra?)

backend/pah_fm/settings.py
```diff
@@ -204,3 +204,6 @@
 JWT_AUTH = {
     'JWT_EXPIRATION_DELTA': datetime.timedelta(days=30),
 }
+
+RSA_PUBLIC_EXPONENT = 257
+RSA_NUMBER_OF_BITS = 19
```
magul: it could probably be named `RSA_BIT_LENGTH` as this method on `int`.

backend/fleet_management/models.py
```diff
@@ -2,20 +2,43 @@
...
-from django.core.validators import MinLengthValidator
+
+from fleet_management.crypto import PublicKey, PrivateKey, find_pair_of_keys
+
 
 def get_current_timestamp_in_gmt():
     return calendar.timegm(time.gmtime())
 
 
+def pad(n: int):
```
magul: Don't really see a lot of reason to have this here as we use `max_length=6` in model's fields so we should to pad to 6.

backend/fleet_management/models.py
```diff
@@ -2,20 +2,43 @@
...
-    rsa_pub_e = models.CharField(max_length=6, validators=[MinLengthValidator(6)], null=False, default='')
-    rsa_priv_d = models.CharField(max_length=6, validators=[MinLengthValidator(6)], null=False, default='')
+    rsa_modulus_n = models.CharField(max_length=6, null=False, default='')
+    rsa_pub_e = models.CharField(max_length=6, null=False, default='')
+    rsa_priv_d = models.CharField(max_length=6, null=False, default='')
+
+    def save(self, regenerate_keys: bool = False, *args, **kwargs):
+        if regenerate_keys or self.pk is None:
+            pub, priv = find_pair_of_keys()
+            self.rsa_modulus_n = pad(pub.n)
```
magul: that should be just `self.rsa_modulus_s = str(pub.n).zfill(6)` and so on...

backend/fleet_management/tests/test_crypto.py
```diff
@@ -0,0 +1,62 @@
...
+
+class CryptoTest(APISimpleTestCase):
+    def setUp(self) -> None:
+        self.n_tests = 1000
+        self.keys = [find_pair_of_keys() for _ in range(self.n_tests)]
+
+    def test_sign_and_verify(self):
+        for pub, priv in self.keys:
+            message = randbits(settings.RSA_NUMBER_OF_BITS)
+            cipher = sign(message, priv)
```
magul: as we name that `sign` I would use `signature` instead of `cipher`.

backend/fleet_management/tests/test_crypto.py
```diff
@@ -0,0 +1,62 @@
...
+    find_prime,
+    find_p_q_phi,
+    find_pair_of_keys,
+)
+
+
+class CryptoTest(APISimpleTestCase):
+    def setUp(self) -> None:
+        self.n_tests = 1000
+        self.keys = [find_pair_of_keys() for _ in range(self.n_tests)]
```
magul: I know 1000 is a small number, but instaed of saving that to the list we could in `test_sign_and_verify` run:
```
for _ in range(self.n_tests):
    pub, priv = find_pair_of_kyes()
   ....
```

backend/fleet_management/crypto.py
arturtamborski: I tried to avoid throwing any exceptions in math functions to keep stuff bit simpler.
Technically prime numbers can be negative which would imply that the -1 could be one of possible outcomes of this function, so maybe the exception would be more sensible, but this function is used only here.

backend/fleet_management/crypto.py
```diff
@@ -0,0 +1,104 @@
...
+    return True
+
+
+def find_prime(bits: int) -> int:
+    """Returns prime number with specified amount of bits."""
+    prime = randbits(bits) | 1
+
+    while True:
+        if is_prime(prime):
+            break
```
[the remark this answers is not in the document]
arturtamborski: I've used `break` so that I wouldn't have multiple return points.

backend/fleet_management/crypto.py
arturtamborski: nice ;)

PR #377, opened 2019-12-06
[310] Script to generate keys for existing users

arturtamborski:
Resolves https://github.com/CodeForPoznan/pah-fm/issues/310, should be merged after #368.


arturtamborski:
Unfortunately, reviewing this won't make much sense until #368 is merged, as it is based on that PR.

PR #378, opened 2019-12-06
[314]  Develop drive hash function

### LINKED ISSUE #314: Develop drive hash function.
The function will accept drive object (maybe in JSON) and will return hash that contains 6 decimal digits (so `mod 10^6`). We cannot use stringify or any other _stdlib magic_, as we need to run exactly the same algorithm in JS and Python and end up with the same result.

arturtamborski:
Closes #314, should be merged after #368.


backend/fleet_management/tests/test_crypto.py
```diff
@@ -10,20 +10,21 @@
...
         self.n_tests = 1000
-        self.keys = [find_pair_of_keys() for _ in range(self.n_tests)]
 
     def test_sign_and_verify(self):
-        for pub, priv in self.keys:
-            message = randbits(settings.RSA_NUMBER_OF_BITS)
-            cipher = sign(message, priv)
-            self.assertTrue(verify(message, cipher, pub))
-            self.assertFalse(verify(message + 1, cipher, pub))
+        for _ in range(1000):
```
Ctilhs: Can you use `self.n_tests` here?

Ctilhs: Couldn't you use `self.n_tests` here?

arturtamborski: I should! good point

PR #393, opened 2019-12-21
Align migrations with master branch

arturtamborski:
### Why
We were not concerned with `master` for few sprints which caused a bit of a drift from current state of database in production environment. I'm proposing these changes as a way to align it safely and automatically with database schema/migrations in `develop`.

### How to test it:
- clean your docker-compose environment including volumes
- checkout to master branch
- start project, generate some random data
- stop the project, checkout `package-lock.json`
- fetch this branch, checkout to it and start the project
- run migrations

Or use this handy one-liner :)
```
make stop && \
docker-compose down --volumes && \
git stash && \
git checkout -- frontend/package-lock.json && \
git checkout upstream/master && \ 
make start && \
sleep 20 && \
make populate-database && \
make stop && \
git checkout -- frontend/package-lock.json && \
git checkout align-migrations-with-master-branch
```

now watch closely the logs - migrations will be applied on start
```
make start logs
```

The result should be a neat output of applied migrations without any errors or exceptions.


backend/fleet_management/models.py
```diff
@@ -19,15 +18,18 @@ class User(AbstractUser):
...
 
-    def save(self, regenerate_keys: bool = False, *args, **kwargs):
-        if regenerate_keys or self.pk is None:
-            pub, priv = find_pair_of_keys()
-            self.rsa_modulus_n = str(pub.n).zfill(6)
-            self.rsa_pub_e = str(pub.e).zfill(6)
-            self.rsa_priv_d = str(priv.d).zfill(6)
+    def save(self, *args, **kwargs):
+        if self.pk is None:
+            self.regenerate_keys()
```
pniedzwiedzinski: This means that `regenerate_keys` is called when user is created?

arturtamborski: Yes, it's called only when user is _created_, not _updated_. We don't want to regenerate keys after every change in their data (though technically we could regenerate it like that, it would still work but there's no point in doing this).

arturtamborski:
Let me describe a bit more clearly what I did here to o give you bigger perspective:

Production DB is in some state. This state has some migrations applied and contains some data but generally it's way behind the develop branch. During all this time we wrote multiple significant migrations (merging Passenger and User) which can be considered major, because they don't just add new column, it but a whole lot more. On top of that we added more changes to DB related to RSA keys.

All of these changes are 'virtual' in a sense that they have not been actually applied to the live DB, which means that they were not tested in terms of compatibility with master branch. They were only tested with previous migration that happened to be there.

So to fix any possible problems that we might encounter when deploying new version I started the DB from master with bunch of randomly generated data and started applying migrations one by one to see how far I could get with them without breaking something, and when it finally happened I patched that migration and went forward. That's how I ended up with these proposed changes.

During this time I spotted few more issues not related strictly to database layout but rather migrations themselves and how they are written down. So I fixed this as well.

In order to confirm that they are now OK one could switch to master, generate some data, switch back to my branch and apply the migrations. They should all pass without any problems, whereas the current state in develop branch fails multiple times.

PR #397, opened 2019-12-28
376 add country column

hkaczmarek90:
Now, should be ok.


backend/fleet_management/api.py
```diff
@@ -27,7 +27,7 @@ class PassengerListView(generics.ListAPIView):
     filter_backends = (filters.OrderingFilter, filters.SearchFilter)
     search_fields = (
         'first_name',
-        'last_name',
```
arturtamborski: I wouldn't remove that, the general rule is that you don't leave trailing commas in inline code:
```
something = [1, 2, 3, 4]
```
but you _do_ leave them when items are listed line-per-line:
```
something = [
    1,
    2,
    3,
    4,
]
```

but that's just nitpick around the style.

arturtamborski:
Good job on that :)

PR #398, opened 2019-12-29
391 display country column in projects

hkaczmarek90:
Added Column and Filter


pniedzwiedzinski:
Isn't #397 the same?

hkaczmarek90:
No.. it's almost the same.
#397 was in Drive model only column. This one is in Project, and here is filter added too

arturtamborski:
Good work on that. Can you resolve the conflict and remove package-lock.json from PR?

pniedzwiedzinski:
> Good work on that. Can you resolve the conflict and remove package-lock.json from PR?

I believe that what @arturtamborski meant was to remove the `package-lock.json` from the diff, not the whole file from the repository. This could help:

```
git checkout upstream/develop -- frontend/package-lock.json
```

backend/fleet_management/models.py
```diff
@@ -57,6 +58,11 @@ def __str__(self):
         return self.title
 
 
+class ProjectAdmin(admin.ModelAdmin):
```
magul: We tend to store derivatives of `ModelAdmin` in `admin.py`, not in `models.py`.

PR #400, opened 2020-01-08
[405] form signature should be treated as optional on backend

### LINKED ISSUE #396: Drive.date DateField should be mandatory so that it can be included in hash computation
`/api/drives` endpoint does not list `date` as a mandatory field which means that it can't be used for computing the drive hash. Frontend side always sends this anyway so we could make it mandatory.

### LINKED ISSUE #405: Form signature should be treated as optional on backend
Javascript code will not send POST request with `signature` field if it was left unfilled which is correct according to the application logic - this field is optional. However, backend does not treat it as such causing it to be always required.
This leads us to bug where frontend keeps resending the same form without the key but backend rejects it.
In order to solve it the backend should treat `signature` as an optional field and use it only when it was provided.

arturtamborski:
Fixes #405  :)


backend/fleet_management/factories.py
```diff
@@ -2,31 +2,65 @@
...
+COUNTRIES = ("UA", "SS")
+
+
+class MakeFactoryMixin:
+    """
+        This class attaches two new static methods to factory class.
 
-COUNTRIES = ('UA', 'SS')
+        make() and make_batch() both overwrite the default behaviour
+        of create() and create_batch() in order to protect the user
```
Ctilhs: Like we discussed on slack. User word here is misleading if you ment developer.

backend/fleet_management/tests/test_cars.py
```diff
@@ -12,10 +12,10 @@
 class CarsApiTestCase(APITestCase):
     def setUp(self):
         self.url = reverse("cars")
-        self.user = UserFactory.create(
+        self.user = UserFactory.make(
             groups=[Group.objects.get(name=Groups.Driver.name)]
         )
-        self.cars = CarFactory.create_batch(size=3, country=self.user.country)
+        self.cars = CarFactory.make_batch(size=3, country=self.user.country)
```
Ctilhs: You can unify how you use make_batch.
`make_batch(size=3)` vs `make_batch(3)`

PR #401, opened 2020-01-11
[375] Add license file

arturtamborski:
Solves #375 
Please review and approve.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Code For Poznań (codeforpoznan.pl)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Code For Poznań (codeforpoznan.pl)
```
arturtamborski:
```suggestion
Copyright (c) 2020 Code For Poznań (codeforpoznan.pl) and contributors.
```

magul:
I will personally try to communicate with all of previous contributors to ensure all are ok with this license applied.

Give me, please, a few weeks to do so.

OtisRed:
Has there been any progress in that regard? Not having a license could be actually a big deal formally, because without it, since we are not PAH volunteers per se, using the app by PAH might constitute the unjustified income. Since there is no other formal agreement between them and us (because "us" don't legally exist), having a license would prevent that problem. I suggest to add one as quick as possible.

@magul I remember we agreed to talk to each contributor and give one a chance to agree. If it's been done and the rest just didn't respond I would take it as a silent consent. If you didn't contact them I would merge it anyway - we all know what we are doing and even if some contributors didn't express that explicitly the implicit working agreement is still binding. I think it's highly unlikely that anyone would litigate and even if somebody were to, the case in that regard seems very frail. Meanwhile from the formal standpoint it seems more pressing to formalize the use of software so that PAH would not be at risk (@OlaPAH please correct me if that's for some reason an irrational panic on my part ^^).

arturtamborski:
bump @OtisRed, @magul, @monikakrawczak any decision on it?

PR #409, opened 2020-01-15
[319] Verify drives on frontend

### LINKED ISSUE #319: Browser-side sign validation based on passenger public key.
When passenger signs received hash value and passes that back to driver he will need to verify the signature using passenger's public key stored in local storage.

arturtamborski:
Resolves #319. This code adds color bar as an indication for drives which were verified.

### DIFF
--- a/backend/fleet_management/serializers.py
+++ b/backend/fleet_management/serializers.py
@@ -16,10 +16,13 @@ class GroupSerializer(serializers.ModelSerializer):
 
 class UserSerializer(serializers.ModelSerializer):
     groups = GroupSerializer(many=True)
+    rsa_modulus_n = fields.CharField(read_only=True)
+    rsa_pub_e = fields.CharField(read_only=True)
+    rsa_priv_d = fields.CharField(read_only=True)
 
     class Meta:
         model = User
-        fields = ("id", "username", "groups")
+        fields = ("id", "username", "groups", "rsa_modulus_n", "rsa_pub_e", "rsa_priv_d")
 
 
 class PassengerSerializer(serializers.ModelSerializer):
--- a/frontend/src/views/DriveFormView.vue
+++ b/frontend/src/views/DriveFormView.vue
@@ -157,9 +157,9 @@
       <label for="driveHash">{{ $t('drive_form.drive_hash') }}</label>
       <input
         id="driveHash"
+        type="number"
         v-model.number="computeHash"
         class="form-control"
-        type="text"
         readonly
       >
     </div>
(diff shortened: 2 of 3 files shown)


frontend/src/views/DriveFormView.vue
```diff
@@ -329,6 +337,7 @@ export default {
     ...mapState(namespaces.passengers, {
       passengers: state =>
         (state.data || []).map(p => ({
+          ...p,
```
pniedzwiedzinski:
```suggestion
          rsaPubE,
          rsaModulusN
```

backend/fleet_management/serializers.py
```diff
@@ -16,10 +16,13 @@ class Meta:
...
 class UserSerializer(serializers.ModelSerializer):
     groups = GroupSerializer(many=True)
+    rsa_modulus_n = fields.CharField(read_only=True)
+    rsa_pub_e = fields.CharField(read_only=True)
+    rsa_priv_d = fields.CharField(read_only=True)
 
     class Meta:
         model = User
-        fields = ("id", "username", "groups")
+        fields = ("id", "username", "groups", "rsa_modulus_n", "rsa_pub_e", "rsa_priv_d")
```
pniedzwiedzinski: Do we need to pass the private key?

arturtamborski: Yes, this one is a different serializer, it's used only for getting the details of currently logged in user (via `/api/users/me`). And we want these details but only for the _logged in_ user.
Passenger serializer defined below (used by Drivers) does not expose the private key.

frontend/src/views/DriveFormView.vue
arturtamborski: I don't quite get how to do it like you suggested. I can't just list `rsaPubE`, `rsaModulusN` like that can I?
I did the usual `rsaPubE: p.rsaPubE`, hope that's what you meant.

pniedzwiedzinski: Right, my bad, maybe something like this:

```js
(state.data || []).map(
  ({id, firstName, lastName, rsaModulusN, rsaPubE}) => ({
          value: id,
          text: [firstName, lastName].join(' '),
          rsaModulusN,
          rsaPubE,
  })),
```

arturtamborski: whoa, that's wild! :D :dancer: Will test it later and see how it goes

arturtamborski: After second thought - don't you think that the current version is good enough?

pniedzwiedzinski: it's ok

PR #418, opened 2020-01-29
Fix gherkin in login feature

kruczyna:
- Fixed Gherkin scenario to comply with BDD principles 
- Reorganized files (POP preparations)
- Added scenario outline for negative login tests
- Did not make any changes to add-new-route-testcase.feature and logout-testcase.feature (files from previous behave tests)


behave_test_automation/pah_behave/features/steps/login_steps.py
```diff
@@ -0,0 +1,52 @@
...
+    password_field = context.driver.find_element(*login_selector.password_field)
+    password_field.click()
+    password_field.send_keys(password_credential)
+
+    login_button = context.driver.find_element(*login_selector.login_button_enabled)
+    login_button.click()
+
+
+@then('User is logged in to pah website successfully')
+def successful_login(context):
```
w1stler: AFAIR functions shouls start with verb.

behave_test_automation/pah_behave/features/page_objects/login_page.feature
```diff
@@ -0,0 +1,21 @@
...
+    In order to be able to login to pah website
+    As a user
+    I want to login only via valid credentials
+
+    Background:
+       Given User navigates to pah-fm website
+
+    Scenario: User successfully logs in to pah-fm system with valid credentials
+       Given User submits login form with "hello@codeforpoznan.pl" login and "cfp123" password
+        Then User is logged in to pah website successfully
```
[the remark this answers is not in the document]
kruczyna: Touché, modified to: `User logs in to pah-fm system`

behave_test_automation/pah_behave/features/page_objects/login_page.feature
```diff
@@ -0,0 +1,21 @@
...
+
+
+    Scenario Outline: User is not able to login to pah-fm with invalid credentials
+       Given User submits login form with "<login>" login and "<password>" password
+        Then User not logged into pah-fm website
```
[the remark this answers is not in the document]
kruczyna: Changed to `User failed to login into pah-fm website`

behave_test_automation/pah_behave/features/steps/login_steps.py
```diff
@@ -0,0 +1,52 @@
+import parse
+from behave import given, then, register_type
+import features.selectors.login_selectors as login_selector
+import features.selectors.base_selectors as base_selector
+from features.steps.common_actions import short_wait
```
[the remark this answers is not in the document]
kruczyna: Changes applied

behave_test_automation/pah_behave/features/page_objects/login_page.feature
```diff
@@ -0,0 +1,21 @@
...
+      Examples:
+        | login                  | password |
```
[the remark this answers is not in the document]
kruczyna: Changes applied

behave_test_automation/pah_behave/features/steps/login_steps.py
kruczyna: Changed to `login_successful` and `login_unsuccessful`

behave_test_automation/pah_behave/features/page_objects/login_page.feature
```diff
@@ -0,0 +1,21 @@
...
+        | 123                    |    123   |
+        | hello@codeforpoznan.pl |          |
+        |                        | cfp123   |
```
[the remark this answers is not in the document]
kruczyna: Separated those cases from each other. Right now there's `User is not able to login to pah-fm with invalid credentials` and `Login button is blocked until user fills out all form fields`

behave_test_automation/pah_behave/features/steps/login_steps.py
```diff
@@ -0,0 +1,61 @@
...
+
+@then('User failed to login into pah-fm website')
+def login_unsuccessful(context):
+    current_page = context.driver.current_url
+    assert current_page == 'http://localhost:8080/login'
+    short_wait(context.driver, login_selector.login_failed_message)
+
+
+@then('Login button is not clickable')
+def login_unsuccessful(context):
```
arturtamborski: would be cool to have this function renamed to something more aligned with statement in `then`.
Or, even better imo - rename all functions with behave decorators (`then`, `when`, `given`) to simple `impl` as in implementation of _that_ step, because we never really call these functions directly, do we?

behave_test_automation/pah_behave/features/steps/login_steps.py
```diff
@@ -0,0 +1,61 @@
...
+    current_page = context.driver.current_url
+    assert current_page == 'http://localhost:8080/login'
```
arturtamborski: I don't think that we need to assure that in every step. I mean, this `Login button is not clickable` will be used in one particular situation, when we know that we are in `/login`, so this check doesn't make much sense.

behave_test_automation/pah_behave/features/steps/login_steps.py
kruczyna: Do you mean something like: `impl_login_button_is_not_clickable` ?

behave_test_automation/pah_behave/features/steps/login_steps.py
kruczyna: Fair enough, but in `User submits login form with "<login>" login and "<password>" password` step I'm click the login button, which is mapped as: `btn-primary` (it's always being found, because I map the id, which is not affected by the button state [disabled attribute]). 
I left this assertion to make sure, that I did not somehow get redirected to the drives page nor anywhere else via the Login button or invalid/empty credentials. 
Does it sound good? :)

behave_test_automation/pah_behave/features/steps/login_steps.py
arturtamborski: not quite, just `def impl(context):` in every function.

behave_test_automation/pah_behave/features/steps/login_steps.py
arturtamborski: Sure :) That's absolutely valid point for me.

behave_test_automation/pah_behave/features/steps/login_steps.py
kruczyna: I will introduce Page Object Pattern in next PR, so I will be calling those functions in different places. With this `def impl(context):` approach it might get messy ;)

arturtamborski: Sure, that's sounds reasonable.

PR #423, opened 2020-02-07
Fix types in form

pniedzwiedzinski:
I was looking through the main form since it gets bigger and bigger after every new feature and I noticed that sometimes we transform types when it's not needed.

This is now a draft since I'm not sure if I found all bugs like this and I will be looking for more.


pniedzwiedzinski:
Ok, I think that's all

PR #427, opened 2020-02-10
261 logout pop up

pniedzwiedzinski:
I added modal that will show the warning to user attempting logout on offline.

I also moved routes to a separate file because there were problem with import, that finally wasn't needed. But I think it's ok to leave it that way.


frontend/src/mixins/GroupGuardMixin.js
```diff
@@ -1,8 +1,8 @@
 import store from '../store';
-import { groupBasedRoutes } from '../router';
+import { groupBasedRoutes } from '../router/routes';
 import { reverseObject } from '../services/reverseObject';
 
-const { user } = store.state;
+const { user } = store ? store.state : { user: undefined };
```
arturtamborski:
```suggestion
const { user } = store && store.state ? store.state : { user: undefined };
```
?

frontend/src/mixins/GroupGuardMixin.js
```diff
@@ -1,8 +1,8 @@
...
 const groups = user ? user.groups.map(({ name }) => name.toLowerCase()) : [];
```
arturtamborski:
```suggestion
const groups = user && user.groups ? user.groups.map(({ name }) => name.toLowerCase()) : [];
```
I mean, it can be undefined, right? If that happens, `map()` will fail.

frontend/src/router/index.js
```diff
@@ -1,112 +1,27 @@
...
-import PassengerSubmitView from '../views/PassengerSubmitView.vue';
+import flatMap from 'array.prototype.flatmap';
 
 import store from '../store';
-import * as mutations from '../store/mutations';
-import { deleteStorageData } from '../services/api/auth';
 import { isUserLoggedIn } from '../services/api/user';
 
-Vue.use(Router);
+import routes, {
```
arturtamborski:
```suggestion
import { routes,
```
Or am I missing something?

frontend/src/store/actions.js
```diff
@@ -55,7 +55,7 @@ export const actions = {
   [LOGOUT]({ commit }) {
     commit(mutations.SET_USER, null);
     deleteStorageData();
-    commit(mutations.SET_LOGOUT_PROGRESS, true);
+    // commit(mutations.SET_LOGOUT_PROGRESS, true);
```
arturtamborski: Is that some leftover from testing?

frontend/src/translations.json
```diff
@@ -243,7 +251,8 @@
     "logout": {
       "thank_you": "Dziękujemy!",
       "successfully_logged_out": "Wylogowano pomyślnie.",
-      "login_again": "Zaloguj się ponownie"
+      "login_again": "Zaloguj się ponownie",
+      "offline_logout_warning": "Nie masz połączenia z internetem, więc jeśli się wylogujesz będziesz mógł się zalogować dopiero jak odzyskasz połączenie."
```
arturtamborski:
```suggestion
      "offline_logout_warning": "Nie masz połączenia z internetem, więc jeśli się teraz wylogujesz to będziesz mógł się zalogować dopiero gdy odzyskasz połączenie."
```

frontend/src/translations.json
```diff
@@ -189,7 +195,9 @@
       "please_correct_errors": "Proszę poprawić następujące błędy:",
       "drives": "Trasy",
       "unexpected_error": "Wystąpił nieoczekiwany błąd. Prosimy o kontakt z administratorem.",
-      "yes": "tak"
+      "yes": "tak",
+      "are_you_sure": "Jesteś pewien?",
```
arturtamborski:
```suggestion
      "are_you_sure": "Jesteś pewien/pewna?",
```
So we're hitting some limitations regarding localisation :P Though we don't store the gender anywhere it still would be nice to be more versatile in this regard, hence my suggestion.

frontend/src/translations.json
```diff
@@ -160,7 +165,8 @@
     "logout": {
       "thank_you": "Thank you!",
       "successfully_logged_out": "Successfully logged out.",
-      "login_again": "Login again"
+      "login_again": "Login again",
+      "offline_logout_warning": "You are offline so if you log out you will no able to log in again until you will be online again"
```
arturtamborski:
```suggestion
      "offline_logout_warning": "You are offline, so if you log out right now then you won't be able to log in until you become online again."
```

frontend/src/router/index.js
pniedzwiedzinski: `routes` are the default export, so it will be outside the brackets.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#Description

frontend/src/store/actions.js
pniedzwiedzinski: I removed logout progress from the store since it's no longer needed

arturtamborski:
I have a question - should this modal be displayed only when user is offline? It's not described as that in the linked issue (although there isn't much detail either), so maybe we should discuss this?

Other thing - I'm not an UX guy, don't know much about that but I thought it would be nice to have the buttons separated from each other, just as we did in DriveFromView with Submit and Reset.
What do you think about it?

frontend/src/mixins/GroupGuardMixin.js
pniedzwiedzinski: Yes, but I did this primarily for testing purposes

The GroupGuardMixin requires the user to be logged in, hence the `user` should be in store. And if user has no groups the `user.groups` should be an empty array

monikakrawczak:
Confirmed with Client, that we need displaying pop up when user is online as well.
I'm opening a new task.

arturtamborski:
No need to, I think that all work can be done in this PR, as it's still open :)

arturtamborski:
In case this was lost somehow: we'd love to display the pop-up developed here no matter the network status (online/offline). Currently it works so it's shown only when user is offline - this can be surprising when the user temporarily gains network connection.

Please modify the PR such that the pop-up always pops up - thanks!

pniedzwiedzinski:
Sorry for the delay but I was unable to get on this. I think in the next meeting I will end this as it sounds easy to implement.

pniedzwiedzinski:
This should be ok now

Issue #395, opened 2019-12-23
Replace 'passengers' POST param with 'passenger'


arturtamborski:
For historical reasons, POST request to `/api/drives` is accepting a `passengers` list containing map with their `id`s. This is no longer valid according to current project requirements - we can safely limit the endpoint so that it accepts single passenger, just like we accept single car or project.

To solve this one would have to take a look into `DriveSerializer` and rewrite `passengers` form field  to `PassengerSerializer` instance with appropriate constraints. More so, it would be nice to also remove the unnecessary mapping in `handleSubmit()` on frontend site.

That's also a good opportunity to write more tests, of course :) Some ideas include checking if request fails with `HTTP_40*` when some fields are not provided, or when some of them have incorrect data and/or type.

arturtamborski:
Also, it would be cool to actually show the Passenger that was assigned to given drive in /drives/ view.

OtisRed:
Might be not relevant but it'll come back if it actually makes troubles

PR #432, opened 2020-02-17
395 replace param passengers with passenger

Ctilhs:
Solves #395


Ctilhs:
@arturtamborski We were discussing it last week.
When working on it I noticed some tweaks which could be done so I created issue #433

arturtamborski:
So I've tested it and seems to work fine but I'm afraid that we'll have to wait with merging it until https://github.com/CodeForPoznan/pah-fm/issues/438 is closed.

PR #435, opened 2020-02-19
Page object pattern to login and logout feature

kruczyna:
Added:
- POP for logout_view.feature
- POP for login_view.feature
- Deleted @skip before_all function in environment.py


behave_test_automation/pah_behave/features/page_objects/logout_page.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class LogoutPage(BasePage):
+    logout_button = (By.CSS_SELECTOR, 'a[href="/logout"]')
+    logout_message = (By.CSS_SELECTOR, '.logout')
+    hamburger_menu = (By.CSS_SELECTOR, 'div.bm-burger-button')
+
+    def logout_via_logout_button(self):
+        self.wait_for_element(self.hamburger_menu)
+        sleep(1)
```
arturtamborski: Why sleep if one line above we `wait_for_element` anyway?

kruczyna: I have major issues with the hamburger_menu button - to be more precise - for waiting until all the events finish loading. None of the selenium wait_for's are handling it properly (staleElement error occurs, or the button is clicked and nothing happens after that [since events are still not loaded]).
I know it's not the best solution, maybe we could make a debug session on that together? For now, I really don't know how to handle it. Maybe you have any ad hoc ideas? :)

arturtamborski: Oh, I see... Selenium is quite stubborn to use when working with Single Page Apps.
I'll look around for some solution to that (it might be needed later on, when more behave tests will be used for the sidebar) but for now on I'll take that as working ;)
Thanks for the explanation.

Issue #442, opened 2020-03-04
Fixture creating user for 'behave tests' needs update.


Ctilhs:
If we have all migrations applied, running `make init-behave` will throw an error.
[Fixture](https://github.com/CodeForPoznan/pah-fm/blob/develop/behave_test_automation/pah_behave/features/data/fixtures.sql) needs an update. 
We made some changes on User model when feature with passenger/driver authorization was implemented.

Ctilhs:
I notice here one problem, but it's not blocking PR fixing it.

We are creating here the user, which is required only for running behave tests.
I think it's related to this issue https://github.com/CodeForPoznan/pah-fm/issues/352

Ctilhs:
Is this user actually used anywhere? :smiley:

Ctilhs:
It looks like this user is not used anymore.

In this case, can we remove `fixtures.sql` and delete it from Makefile?

arturtamborski:
sure, that sounds even better

Ctilhs:
Done https://github.com/CodeForPoznan/pah-fm/pull/443

PR #443, opened 2020-03-04
442 Update SQL statement.

Ctilhs:
Fix https://github.com/CodeForPoznan/pah-fm/issues/442

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,9 +80,6 @@ debug-backend:  ## Debug backend container (Django)
 populate-database:  ## Populate database with factory based data
 	make manage CMD=populate_database
 
-init-behave:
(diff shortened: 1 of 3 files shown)


Ctilhs:
Waiting for answer here https://github.com/CodeForPoznan/pah-fm/issues/442

arturtamborski:
The numbers you've inserted there for rsa key are incorrect. They are not prime and the public exponent differs from the one defined in settings which would mean that every computation, even with correct key, would fail.

I much rather would like to see a proper rsa key there, but that is not needed anyway, because why would we want raw sql queries here to update stuff in the table if we have python, models and factories :)

Anyway, go on and remove this file along with its directory and mention in Readme. It's not needed anymore, we have `make populate-database` for that purpose.

Ctilhs:
Good point with RSA elements :+1: 

I removed redundant stuff.

PR #446, opened 2020-03-04
444 issue fix

kruczyna:
Changed path in README.md description from 
`~/codeforpoznan/pah-fm/pah_behave`
to:
`~/codeforpoznan/pah-fm/behave_test_automation/pah_behave`


arturtamborski:
That works fine, thanks. Please work a bit on the git flow approach, because this simple change shouldn't require 8 commits. You've probably forgot to checkout from clean develop?

PR #449, opened 2020-03-04
436 add filter in django

hkaczmarek90:


backend/fleet_management/models.py
```diff
@@ -51,6 +51,11 @@ def __str__(self):
         return self.plates
 
 
+class CarAdmin(admin.ModelAdmin):
```
magul: just a side note here - usually whatever related to Django admin is stored in `admin.py` file (not in `models.py`)

PR #454, opened 2020-03-10
Drive verification in separate view

### LINKED ISSUE #424: Verify drive after submit
Currently, the drive signature is a part of the drive form, so it means that we can alter the drive details after verification. Of course, this means that the drive is unverified since the checksum also changes, but recently @magul pointed out that the drive could be verified after clicking the 'submit' button. Something like this:

image
image

(I used `passenger-input` class so the text is bigger)

pniedzwiedzinski:
This will close #424 

This is  working by now, but I think that in the future it would be good to redirect after submit to `/drive` route and display notifications there.


arturtamborski:
I think that the second view (with verification code) should redirect the user to home page after submit.
Currently it does not, which allows me to send the same form with different codes multiple times.

pniedzwiedzinski:
I created this temporary solution that will hide the form and display the notifications with a button that will redirect to the form.

arturtamborski:
Fine by me, we can improve it later on. One question - are you planning on indicating that the signature is optional (as commented here: https://github.com/CodeForPoznan/pah-fm/issues/424#issuecomment-583752617) or you think that should be done next time we touch this form?
I'm fine with both options :)

frontend/src/views/DriveVerifyView.vue
```diff
@@ -0,0 +1,172 @@
+<template>
+  <div>
+    <main-form
+      v-if="!submitted"
+      :title="$t('common.confirm_drive')"
```
arturtamborski:
```suggestion
      :title="{{ $t('common.confirm_drive') }}"
```
or am I missing something?

pniedzwiedzinski: `:title="$t(...)"` is equivalent to `title="{{ $t(...) }}"`, there is a colon before `title`

pniedzwiedzinski:
> Fine by me, we can improve it later on. One question - are you planning on indicating that the signature is optional (as commented here: [#424 (comment)](https://github.com/CodeForPoznan/pah-fm/issues/424#issuecomment-583752617)) or you think that should be done next time we touch this form?
> I'm fine with both options :)

Oh I forgot about this. I will add it now.

PR #453, opened 2020-03-06
Limit avaiable countries to just a few (fixes 448)

arturtamborski:
image

PR #455, opened 2020-03-11
Filter by used country only

hkaczmarek90:
My solution for Issue #453


arturtamborski:
Originally proposed here, but merged in https://github.com/CodeForPoznan/pah-fm/pull/469.

PR #471, opened 2020-03-21
Add new drive scenarios

kruczyna:
Scenarios added:
- User sees field validation errors on add new drive view
- User submits a new drive


arturtamborski:
please resolve above conflicts :)

PR #490, opened 2020-04-19
461 Detect when user was last time online

bckamil:


backend/fleet_management/migrations/0023_user_last_seen.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('fleet_management', '0022_auto_20191221_2303'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='last_seen',
+            field=models.DateTimeField(blank=True, null=True),
```
arturtamborski, line 16:
```suggestion
            field=models.DateTimeField(blank=True, null=False),
```
I suggest avoiding nulls overall in database design. `blank` works fine where we don't care that much about the value.

backend/fleet_management/models.py
```diff
@@ -17,6 +17,7 @@ class User(AbstractUser):
     rsa_modulus_n = models.CharField(max_length=6, null=False, default="")
     rsa_pub_e = models.CharField(max_length=6, null=False, default="")
     rsa_priv_d = models.CharField(max_length=6, null=False, default="")
+    last_seen = models.DateTimeField(null=True, blank=True)
```
arturtamborski, line 20:
```suggestion
    last_seen = models.DateTimeField(null=False, blank=True)
```
see comment above

backend/fleet_management/middleware.py
```diff
@@ -0,0 +1,10 @@
+from django.utils import timezone
+from django.utils.deprecation import MiddlewareMixin
+
+from fleet_management.models import User
+
+
+class UpdateLastSeenMiddleware(MiddlewareMixin):
+    def process_request(self, request):
```
arturtamborski, line 8: shouldn't that be `__call__()`? Because that looks like a Pre 1.10 code according to docs:
https://docs.djangoproject.com/en/2.2/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

Issue #459, opened 2020-03-18
Move everything from localStorage to vuex store


arturtamborski:
We directly use localStorage in Drive Form View and Save Token (and more?).
We don't have to worry about the values in there for now.

pniedzwiedzinski:
There's a problem with binding input fields with v-model to vuex. There's a package that resolves this issue: https://markus.oberlehner.net/blog/form-fields-two-way-data-binding-and-vuex/
I will test it now.

pniedzwiedzinski:
Looks ok https://codesandbox.io/s/recursing-frog-6rsrw

pniedzwiedzinski:
We need to also move the jwt token into the store.

Issue #421, opened 2020-02-07
Vuex modules


pniedzwiedzinski:
We use vuex as our store management, and our store is BIG. I don't really know how it's working 😄.

We could split the store into smaller modules, that will run independently from each other. This hopefully would make the store easier to maintain, test, and in future this might also help with "store migrations".

We actually use modules now, but we store all the methods in common files. So this shouldn't be hard to change

pniedzwiedzinski:
Actually we need store migrations to perform this 😄

pniedzwiedzinski:
I've been thinking about this recently and I came to following. I split the store by data destination and purpose. Thanks to that we will have state and actions divided in category which hopefully would be easier for newcomers to comprehend. Also I want for this module to be independent from each other, so it will be easier to interact.

> NOTE: I put `?` next to the drive_form since I'm not sure where to store it. On the one hand it's the current state of the application, but on the other hand it might contain data that we don't want to lose.

## Modules

### API module

This would store responses from the server so when the network is down the app would still have access to the data. This data is not sensitive and we don't really care about it hence all of this is backed up on the server. I also want to get rid of the `error` field since we don't use it anywhere and maybe we should start thinking about notification system.

- `cars`
- `passengers`
- `drives`
- `projects`

### Session module

This would store the data about current user and tokens. It would be nice to keep the session during the update process.

- `user`
- `token`
- `language`

### Data module

The most sensitive! There will be stored all data that needs to be synced with server. Most of the migration would be related to this module.

- `unsynced_drives`
- `incorrect_drives`
- `drive_form`?

### State module

This would cover the communication between the components of the app. The data in this module wouldn't be persistent so we don't really have to bother about it.

- `drive_form`?
- `drive_hash`
- `isOnline`
- `hash`
- `loginInProgress`

### Migration module [NEW✨]

This would keep track of applied migration to the store and check for the application update.

- `applied_migrations` (#460)
- `current_version` (#462)
- `updateReady`

arturtamborski:
That looks really good!

Regarding `drive_form` - I think that it should belong to Data module.
My reasoning here is that we can treat the partial data typed into the form (and selected car plates) as some sort of the user settings that one can have control over.
Maybe it's not the most _crucial_ part of the data that we store, but it's still one of the main points that user will interact with, so we should have a way of storing it, as opposed to keeping it within State module (which holds temporary information that is not _that_ important).

[...]

PR #492, opened 2020-05-07
PAH-470 update dependencies

juliaanholcer:
- Update dependencies.

Note: "stylelint-processor-html@1.0.0" is deprecated and it seems that we might not use the scss preprocessor for stylelint, because its latest version should be able to handle .vue files, however, removing this package or replacing it with non-deprecated preprocessor will cause that style linter doesn't work at all. Sorry, but I have no idea how to solve it so this package stays.

@arturtamborski I'm not quite sure which dependencies you meant as unused.


juliaanholcer:
Haha, it seems that now eslint works and there is a lot of lint errors in the code ;-)

arturtamborski:
> Sorry, but I have no idea how to solve it so this package stays.
unfortunately I can't really help you here. I guess it's not that big of a deal, certainly not compared to all other packages updated/removed :)

> Haha, it seems that now eslint works and there is a lot of lint errors in the code ;-)
yes, but there's a simple way to fix it:
```
make bash-frontend
npm run lint:fix
```
aaand `git add .` :) there will still be 11 errors that need to be fixed by hand, but that's a lot less than ~80 originally reported.

juliaanholcer:
Remaining problems arise from dependency cycles, i.e. auth.js imports some stuff from http.js and vice versa. One can take shortcuts and disable eslint for these imports, but we may need to check if such code is optimal. Unfortunately, many things in the current vershion are new to me. @pniedzwiedzinski , do you have an idea of how to refactor to avoid these cyclical dependencies?

pniedzwiedzinski:
We could divide `auth.js` into 2 parts:
- interacting with `localStorage` (`getToken`, `getAuthHeader`)
- interacting with api (that's `http.js` is imported for)

But as in #459  (#421) we will remove localStorage, so this is a temporary fix.

pniedzwiedzinski:
This should fix it: [circular-import.patch.txt](https://github.com/CodeForPoznan/pah-fm/files/4610332/circular-import.patch.txt)

Run `git apply /path/to/circular-import.patch.txt` in pah-fm root dir

pniedzwiedzinski:
@juliaanholcer what's the status on that?

juliaanholcer:
> @juliaanholcer what's the status on that?

I'm going to finish it but unfortunately not this week, sorry.

pniedzwiedzinski (approved):

arturtamborski:
Will reopen with new updates, this one is too old to bother fixing.

PR #494, opened 2020-05-13
[483] Add Somali translation

### LINKED ISSUE #483: Somali translation
Add Somali translation option for Application (driver and passenger).

For a driver, a passenger with a country defined as Somalia set the Somali language as default.
Users can always change the language.

arturtamborski:
Closes #483


pniedzwiedzinski:
What about lower and upper casing?
image
It this how this language works? :laughing:

arturtamborski:
I really have no idea. Most of this translation was handed over, so I just assumed that it is correct.
I'm assuming that we will get some feedback from the users so we can fix the errors afterwards.

PR #497, opened 2020-05-17
[406] Add tests for /api/passengers

arturtamborski:


backend/fleet_management/tests/test_passengers.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_get_all_passengers(self):
+        self.client.force_login(self.user)
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(self.passengers), len(res.data))
+        for idx, passenger in enumerate(sorted(res.data, key=lambda p: p["id"])):
+            self.assertEqual(passenger["id"], self.passengers[idx].id)
+            self.assertEqual(passenger["first_name"], self.passengers[idx].first_name)
+            self.assertEqual(passenger["last_name"], self.passengers[idx].last_name)
```
pniedzwiedzinski, line 34: What about keys? They are also returned on that endpoint

arturtamborski: good point, should check em as well

arturtamborski: aaaaand here's the update :) sorry for delay and please review :)

backend/fleet_management/tests/test_passengers.py
```diff
@@ -0,0 +1,44 @@
...
+            self.assertEqual(passenger["rsa_pub_e"], self.passengers[idx].rsa_pub_e)
+            self.assertEqual(passenger["rsa_modulus_n"], self.passengers[idx].rsa_modulus_n)
+
+    def test_search_passengers_by_first_name(self):
+        self.client.force_login(self.user)
+        url_params = urlencode({"search": self.passengers[0].first_name})
+        res = self.client.get(f"{self.url}?{url_params}")
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(res.data), 1)
```
magul, line 43: It will fail if in the (rare) occasion we will get multiple passengers with the same first name

arturtamborski: Can't believe but it actually happened on travis, on master branch...
https://travis-ci.com/github/CodeForPoznan/pah-fm/builds/176762350

PR #500, opened 2020-05-19
[464] Bump Python to 3.8

### LINKED ISSUE #464: Bump python to 3.8
yay!

arturtamborski:
Closes #464


backend/requirements/base.txt
```diff
@@ -6,7 +6,8 @@ djangorestframework-camel-case==1.0b1
 djangorestframework-jwt==1.11.0
 djangorestframework==3.9.1
 Markdown==3.0.1
-psycopg2==2.7.5
 Pygments==2.3.1
 django-countries==5.3.3
 django-import-export==2.0.2
+serverless-wsgi==1.7.4
```
magul, line 12: are these changes required to bump python?

docker-compose.yml
```diff
@@ -16,7 +16,9 @@ services:
 
   backend:
     networks: [pah]
-    image: codeforpoznan/pah-fm-backend
+    build:
+      context: ./backend
+    #image: codeforpoznan/pah-fm-backend
```
magul, line 21: As I understand by default we want to pull image from docker hub, not building it locally.

backend/requirements/base.txt
arturtamborski: nope, that slipped trough from other pull request

docker-compose.yml
arturtamborski: yes, will correct that

arturtamborski:
> We also need to remember to:
> 
>     * upgrade it on production environment (which will not be so easy, I guess)
> 
>     * upgrade that in dev lambdas

absolutely, that's why I'm marking this one as a draft. It still needs polishing as you already spotted. Also, the live migration will be quite tricky indeed so it might be easier to move the prod environment completely to lambdas instead of bumping it there.

Anyway, that's a draft, please don't mind it too much :)

arturtamborski:
Few notes from me when this finally gets its time to shine:
1. It would be nice to test this PR again before merging it. It works at the moment but who knows what can change in few months later.
2. Dev environment needs to be updated as well, which will boil down to bumping the language versions in lambda definitions in infra repo (2 line change).
3. Production needs to be updated, probably manually.

For now I'm leaving it as is.

PR #507, opened 2020-05-20
Jwt to vuex (http module)

### LINKED ISSUE #459: Move everything from localStorage to vuex store

pniedzwiedzinski:
This will close #459 

I removed the patch method since we don't use it anymore and I don't think we will in the future.


pniedzwiedzinski:
My tests look over-complicated :disappointed:, on the other hand http mocks + vuex mocks doesn't sound easy

pniedzwiedzinski:
I don't have idea how to test token expiration, that's why there're 2 possible outcomes of the last test.

PR #519, opened 2020-06-24
[388] Format backend with black, add hooks to Makefile and travis

arturtamborski:
:sparkle: finally!


Makefile
```diff
@@ -66,6 +69,9 @@ lint-backend:  ## Run linters on backend container
 lint-frontend:  ## Run linters on frontend container
 	docker-compose run --rm --no-deps frontend npm run lint:fix
 
+format-backend:  ## Run formatters on backend
```
pniedzwiedzinski, line 72: `format` vs `lint` - this is inconsistent, we should use one name for frontend and backend

arturtamborski: totally agree, what I wanted to do here is to split the commands into three categories - `lint`, `format`, `test`, where `lint-*` checks for code style with complex rules (`eslint`/`flake8`), `format-*` checks for unified code style (lower level than linter, `@vue/cli-plugin-eslint`/`black`) and `test-*`, well, tests (jest/pytest).

Though that might be bit too overcomplicated? I mean, I can merge the `format-backend` with `lint-backend` and leave it like that.

pniedzwiedzinski: I don't see when would we need separate `lint` and `format` - I would merge it.

arturtamborski: okay, will do

backend/fleet_management/api.py
```diff
@@ -14,59 +14,48 @@
...
 
     def get_serializer_context(self):
-        return {'driver': self.request.user}
+        return {"driver": self.request.user}
 
     def get_queryset(self):
-        return Drive.objects.filter(
-            driver__id=self.request.user.id,
-        )
+        return Drive.objects.filter(driver__id=self.request.user.id,)
```
magul, line 58: that's unconnected with this pull request, but I'm pretty sure we could write that as:
```python
        return Drive.objects.filter(driver=self.request.user)
```

backend/fleet_management/api.py
```diff
@@ -76,6 +65,4 @@ class ProjectView(generics.ListAPIView):
     serializer_class = ProjectSerializer
 
     def get_queryset(self):
-        return Project.objects.filter(
-            country=self.request.user.country,
-        )
+        return Project.objects.filter(country=self.request.user.country,)
```
magul, line 68: it's also super strange, that it left this trailing `,` inside function arguments

magul:
Do you want to review it @pniedzwiedzinski ? I will wait for a week and if there will be no activity from your side I will ignore your (already outdated) review and will merge it.

backend/fleet_management/api.py
arturtamborski: https://github.com/psf/black/issues/1288

PR #523, opened 2020-07-08
[482] Refuel model

w1stler:
There are some issues to be fixed - from my initial work & observations:
- page doesn't work on Safari :E
- requirements are not updated when creating new image
- users are not created (at least commands from `docker_entrypoint.sh`)
- make populate database doesn't work with new import due to the second point probably (need to investigate a bit)


arturtamborski:
>     * page doesn't work on Safari :E

Linking with https://github.com/CodeForPoznan/pah-fm/issues/537 for reference. BTW any update on that?


>     * requirements are not updated when creating new image

Another reference: https://github.com/CodeForPoznan/pah-fm/issues/529


>     * users are not created (at least commands from `docker_entrypoint.sh`)

Can you elaborate on that? I never encountered such problem.


>     * make populate database doesn't work with new import due to the second point probably (need to investigate a bit)

That might be true but without any stack trace/error message its hard to tell.

backend/fleet_management/models.py
```diff
@@ -125,3 +127,17 @@ def hash_form(initial_data: dict) -> int:
...
+
+class Refuel(models.Model):
+    driver = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="drives_refueled"
+    )
+    car = models.ForeignKey(Car, null=False, on_delete=models.CASCADE)
+    date = models.DateField(auto_now_add=True, blank=False)
+    current_mileage = models.IntegerField(null=False)
+    refueled_liters = models.IntegerField(null=False)
+    price_per_liter = models.IntegerField(null=False)
```
arturtamborski, line 140:
```suggestion
    current_mileage = models.PositiveIntegerField(null=False)
    refueled_liters = models.PositiveIntegerField(null=False)
    price_per_liter = models.PositiveIntegerField(null=False)
```
https://docs.djangoproject.com/en/3.0/ref/models/fields/#positiveintegerfield

None of the above fields can be smaller than zero but someone could input them as such in Django admin, which could mess data exports a bit.

backend/fleet_management/models.py
```diff
@@ -125,3 +127,17 @@ def hash_form(initial_data: dict) -> int:
...
+    currency = MoneyField(
```
arturtamborski, line 141:
```suggestion
    price = MoneyField(
```
Seems more appropriate considering that we're tracking how much someone have spent during refuel.

magul: `cost` or `total_cost`?

magul: another thing - do we need to store `cost` (`currencty` currently), `price_per_liter` and `refueled_liters`, as it should be easily calculable as
```
cost (that probably could be <refuel_cost> = price_per_liter * refueled_liters
```

backend/fleet_management/models.py
arturtamborski, line 140:
```suggestion
@property
def price_per_liter(self):
    # TODO: add some checks which would wrap returned value to 2 decimal places
    return self.currency / self.refueled_liters
```
`price_per_liter` can be calculated from `refueled_liters` and `currency` fields.

I know that it was mentioned in ticket but I'm not sure if we really have to have separate column when we can derive it from other fields. That field would have to be filled by user which matters, when you have to fill lengthy form vs short 2-3 field form.

backend/fleet_management/tests/test_refuels.py
```diff
@@ -0,0 +1,25 @@
...
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APITestCase
+from fleet_management.constants import Groups
+from fleet_management.factories import UserFactory
+
+
+class RefuelsApiTestCase(APITestCase):
+    def setUp(self):
+        self.url = reverse("projects")
```
arturtamborski, line 12: you're fetching from `/projects` ?

backend/fleet_management/tests/test_refuels.py
```diff
@@ -0,0 +1,25 @@
...
+        self.user = UserFactory.create(
+            groups=[Group.objects.get(name=Groups.Driver.name)]
+        )
+
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_get_all_refuels(self):
```
arturtamborski, line 21: not very useful without any refuel in DB

backend/fleet_management/serializers.py
```diff
@@ -164,3 +164,21 @@ def is_valid(self, raise_exception=False):
             ):
                 err.status_code = status.HTTP_409_CONFLICT
             raise err
+
+
+class RefuelSerializer(serializers.ModelSerializer):
+    driver = UserSerializer(read_only=True)
+    car = CarSerializer()
```
arturtamborski, line 171:
```suggestion
```

and add that in Meta - https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

backend/fleet_management/serializers.py
```diff
@@ -164,3 +164,21 @@ def is_valid(self, raise_exception=False):
...
+        fields = (
+            "id",
+            "driver",
+            "car",
+            "date",
+            "current_mileage",
+            "refueled_liters",
+            "price_per_liter",
+            "currency",
+        )
```
arturtamborski, line 184:
```suggestion
        )
        read_only_fields = ("driver", "car")
```

backend/fleet_management/tests/test_refuels.py
```diff
@@ -0,0 +1,25 @@
...
+    def setUp(self):
+        self.url = reverse("projects")
+        self.user = UserFactory.create(
+            groups=[Group.objects.get(name=Groups.Driver.name)]
+        )
+
+    def test_401_for_unlogged_user(self):
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_401_UNAUTHORIZED)
+
```
arturtamborski, line 20:
```suggestion
    def test_can_create_a_refuel(self):
        payload = {
            ...
        }

        self.client.force_login(self.driver)
        res = self.client.post(self.url, data=payload, format="json")

        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
        refuel = Refuel.objects.filter(id=res.data["id"])
        self.assertEqual(refuel.count(), 1)
        self.assertTrue(refuel.currency, "???")
        ...
```

The endpoint should support POST and GET, but there's no test for POST.

backend/fleet_management/admin.py
```diff
@@ -126,3 +126,17 @@ class CustomUserAdmin(UserAdmin):
         ),
         (_("Important dates"), {"fields": ("last_seen", "last_login", "date_joined")}),
     )
+
+
+@admin.register(Refuel)
+class RefuelAdmin(admin.ModelAdmin):
+    list_filter = ("driver", "car")  # project, country?
```
arturtamborski, line 133:
```suggestion
    list_filter = ("driver", "car", "project", CountryFilter)
```

AC3: There should be filtering by country, project, driver, and car enabled.

backend/fleet_management/models.py
arturtamborski: Ha! so we had the same idea but going in two different directions!
You wanted to calculate total refuel cost whereas I wanted to calculate price per liter (https://github.com/CodeForPoznan/pah-fm/pull/523#discussion_r464920023)

Which one do you prefer?

magul: yes, we have the same idea, but I don't prefer either of direction here.

I believe that should be confirmed with @OlaPAH but probably drivers would fill total cost and litres refuelled and we will calculate cost per litre.

backend/fleet_management/tests/test_refuels.py
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        self.client.force_login(self.user)
+        res = self.client.post(self.url, data=payload, format="json")
+        self.assertEqual(res.status_code, status.HTTP_201_CREATED)
+
+        refuel = Refuel.objects.filter(id=res.data["id"])
+        self.assertEqual(refuel.count(), 1)
+
+        refuel = refuel[0]
```
arturtamborski, line 66:
```suggestion
        refuel = Refuel.objects.get(id=res.data["id"])
        self.assertEqual(refuel.count(), 1)

```
why not that? you're fetching the record by `id` anyway, so it's ensured by DB that you won't find two records with same id. Also, the `refuel = refuel[0]` won't be needed then.

arturtamborski (changes requested):
I'm sorry for such late review. I've pointed out a few things to consider.

backend/fleet_management/management/commands/populate_database.py
```diff
@@ -106,6 +107,9 @@ def info(message):
                 driver=random.choice(drivers),
                 car=random.choice(cars),
             )
+        info("Creating 50 random refuels")
+        for _ in tqdm(range(50)):
+            RefuelFactory.create()
```
arturtamborski, line 112:
```suggestion
            RefuelFactory.create(driver=random.choice(drivers), car=random.choice(cars))
```
without that `driver=` and `car=` the RefuelFactory will create new users and cars instead of using the already generated ones.

arturtamborski: bump, that is still an issue to correct

backend/fleet_management/serializers.py
```diff
@@ -164,3 +165,41 @@ def is_valid(self, raise_exception=False):
...
+            driver = User.objects.get(id=data["driver"]["id"])
+            car = Car.objects.get(id=car_id)
+        except ObjectDoesNotExist as e:
+            raise ValidationError(e.args[0])
+
+        total_cost = Money(
+            currency=data["total_cost.currency"], amount=data["total_cost.amount"]
+        )
+
+        with transaction.atomic():
```
arturtamborski, line 200: why is it wrapped with transaction? you don't seem to create any other object other than Refuel so it seems to me a bit unnecessary.

backend/fleet_management/tests/test_refuels.py
arturtamborski: another bump. I don't want to 'force' my changes but I'd like to hear your reasoning behind it.

arturtamborski (changes requested):
I only require a change for the first comment (the one with generating models), because it will quickly create lots of users for no reason.
Two other comments are minor but still I'd like to hear a feedback.

backend/fleet_management/tests/test_refuels.py
w1stler: I was looking at convention of previous tests and in drives we have:

```        self.assertEqual(drive.count(), 1)

        drive = drive[0]
        self.assertTrue(drive.is_verified)```

File: backend/fleet_management/management/commands/populate_database.py (hunk 10 above)
Comment (w1stler) [reply backend/fleet_management/management/commands/populate_database.py:112]: Done

File: backend/fleet_management/serializers.py (hunk 11 above)
Comment (w1stler) [reply backend/fleet_management/serializers.py:200]: Also I was looking at drive model:

```
        with transaction.atomic():
            drive = Drive.objects.create(
                **validated_data,
                is_verified=is_verified,
                driver=self.context["driver"],
                car=car,
                project=project,
                passenger=passenger
            )
            drive.save()

            return drive
```

arturtamborski (approved):

PR #524, opened 2020-07-08
Fix login error, await for promises

### LINKED ISSUE #521: Fix the error that makes it impossible to login
Cherry-picked to this commit: https://github.com/CodeForPoznan/pah-fm/commit/cb49a583b59b03b7c2e20578944deef4238b2519
Probably something with missing await. Can be observed on https://dev.pahfm.codeforpoznan.pl

arturtamborski:
Fixes #521

### DIFF
--- a/frontend/src/store/modules/http/state.js
+++ b/frontend/src/store/modules/http/state.js
@@ -1,1 +1,7 @@
 export const TOKEN = 'TOKEN';
+
+export default {
+    // LEGACY SUPPORT
+    // After merge to new version remove localStorage
+    [TOKEN]: localStorage.getItem('jwt') || null,
+};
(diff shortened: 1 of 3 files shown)


frontend/src/store/modules/http/state.js
```diff
@@ -1,8 +1,7 @@
 export const TOKEN = 'TOKEN';
 
 export default {
-  // LEGACY SUPPORT
-  // After merge to new version remove localStorage
-  [TOKEN]: localStorage.getItem('jwt') || null,
+    // LEGACY SUPPORT
```
magul, line 4: that's just nitpicking, but in JavaScript world, indentation is usually multiplication of 2.

arturtamborski: Yeah, I know, I actually removed that code but then reverted it with wrong indentation, hence the change. Will fix it

PR #535, opened 2020-07-21
Limit amount of records shown to Driver in Drives View

### LINKED ISSUE #522: Limit amount of records shown to Driver in Drives View
Select records from last 30 days.

mateuszmagulski:
fixes #522

### DIFF
--- a/backend/fleet_management/api.py
+++ b/backend/fleet_management/api.py
@@ -59,8 +59,7 @@ class DriveView(generics.ListCreateAPIView):
 
     def get_queryset(self):
         return Drive.objects.filter(
-            driver=self.request.user,
-            date__gte=timezone.now() - timedelta(days=30)
+            driver=self.request.user, date__gte=timezone.now() - timedelta(days=30)
             )
 
 
--- a/backend/fleet_management/api.py
+++ b/backend/fleet_management/api.py
@@ -60,7 +60,7 @@ class DriveView(generics.ListCreateAPIView):
     def get_queryset(self):
         return Drive.objects.filter(
             driver=self.request.user, date__gte=timezone.now() - timedelta(days=30)
-            )
+        )
 
 
 class ProjectView(generics.ListAPIView):
--- a/backend/fleet_management/api.py
+++ b/backend/fleet_management/api.py
@@ -1,3 +1,4 @@
+from datetime import date, timedelta
 from rest_framework import generics, filters, views
 from rest_framework.response import Response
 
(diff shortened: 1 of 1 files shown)


backend/fleet_management/api.py
```diff
@@ -1,3 +1,4 @@
+from datetime import date, timedelta
```
arturtamborski, line 1:
```suggestion
from datetime import timedelta

from django.utils import timezone
```

backend/fleet_management/api.py
```diff
@@ -55,7 +56,10 @@ def get_serializer_context(self):
         return {"driver": self.request.user}
 
     def get_queryset(self):
-        return Drive.objects.filter(driver=self.request.user)
+        return Drive.objects.filter(
+            driver=self.request.user,
+            date__gte=date.today() - timedelta(days=30)
```
arturtamborski, line 61:
```suggestion
            date__gte=timezone.now() - timedelta(days=30)
```

backend/fleet_management/api.py
```diff
@@ -55,7 +58,10 @@ def get_serializer_context(self):
         return {"driver": self.request.user}
 
     def get_queryset(self):
-        return Drive.objects.filter(driver=self.request.user)
+        return Drive.objects.filter(
+            driver=self.request.user,
+            date__gte=timezone.now() - timedelta(days=30)
+            )
```
arturtamborski, line 64:
```suggestion
        return Drive.objects.filter(
            driver=self.request.user, date__gte=timezone.now() - timedelta(days=30)
        )
```

arturtamborski:
> As long as we use DateField for date in Drive model, should I use timezone.now().date() or it doesn't matter?

```
In [6]: timezone.now() - timedelta(days=30)
Out[6]: datetime.datetime(2020, 6, 21, 12, 56, 10, 89948, tzinfo=<UTC>)

In [7]: timezone.now().date() - timedelta(days=30)
Out[7]: datetime.date(2020, 6, 21)
```

if you'd use `timezone.now().date()` then you'd lost the `tzinfo` param, so the timezone fiddling wouldn't matter anyway.
In other words you shouldn't use `.date()` because you'd lost the advantage of timezone-aware `datetime` object.

The comparison (`date__gte=`) will still work because whether or not the `datetime` object is aware of timezone doesn't matter, Python still knows how to compare them.

arturtamborski:
Damn, my proposal with black wasn't perfect so it failed, sorry. You'll have to run black locally again and send fix.

PR #544, opened 2020-08-19
Sync button [WIP]

OtisRed:
Two things are happening here:

1. The red/green dot in the top left corner can be replaced with the offline icon next to the menu. It's not necessary - it just came up while designing sync button. So the question is if from the driver perspective it's better or not? @OlaPAH 

Suggested change looks like this:
Zrzut ekranu z 2020-09-09 20-38-01

2. [Work in progress] The sync button at this point is limited to syncing drives and it should be seen only on the rare occasion when: the user is online and there are unsynced drives (so that it's there only when automated syncing fails).

Few more things to do here
* Limit the button visibility only to the /drves path when there is sth to sync - done
* Fix css classes (make some distance between icons on smaller screens)
* Define fail somehow and render it through icon
* Make animations so that the changes in the button aren't so sharp
* Resolve conflicts and clean up the code


arturtamborski:
I'm sorry for not explaining this in more detail in the ticket because it seems to me that you clearly lacked the context when developing this ticket.

> How do you feel about this solution, visually?

What I had in mind was a little indicator/button that would non-intrusively allow for resyncing of the data on demand.
Something like a small button with a refresh icon that upon clicking would spin around few times to indicate a progress of requesting the data and in the end would turn back into a red/green refresh icon. That in itself would indicate whether or not the request was successful. I would stay away from implementing something more sophisticated (even something as simple as showing the message error) because the whole feature itself was supposed to be a side-thing that user could use to force an action or to give him back some of the control over his drives.
The app itself does already sync drives by itself and it does it every minute automatically depending on internet connection available. This sync button would be a second resort option when the regular sync didn't work for some reason.
I'm sorry for not stating it more clearly in the ticket.

> Should sync action return some kind of message like "syncing was successfull" etc.?

I think no, it shouldn't because that's a lot more work (translations) for little profit. What could we show to the user to help him resolve the possible error? Indicate that he has no internet connection? If that's the case then I'd block the user from clicking the button in the first place by greying it out.

frontend/src/App.vue
```diff
@@ -24,113 +34,129 @@
...
-      /* eslint-disable-next-line no-underscore-dangle */
-      this._i18n.locale = this.language;
-    }
-    if (this[IS_USER_LOGGED_IN]) {
-      this.sync();
-      this.fetchUser();
-    }
-  },
-};
+  import { mapState, mapActions, mapGetters } from 'vuex';
```
arturtamborski, line 37: I agree that it makes sense to indent the code in `<script>` and `<style>` but we are not doing it in this project (at least for now) so if you'd like to see that, we can introduce a formatter with such rule, but generally I'd stay away from reformatting the code against existing (although unwritten) rules.

frontend/src/components/Status.vue
```diff
@@ -1,52 +1,64 @@
 <template>
-  <div
-    v-if="!isLogin && user"
-    class="is-online"
-    :class="{ online: isOnline }"
-  />
+  <div v-if="!isLogin && user">
+    <div v-if="isOnline" class="online" @click="">
```
arturtamborski, line 3:
```suggestion
    <div v-if="isOnline" class="online">
```
or am I missing something?

frontend/src/App.vue
```diff
@@ -1,21 +1,31 @@
...
+        </div>
+        <div class="col-3">
+          <Status />
+        </div>
+        <div class="col-3">
+          <sidebar :show="showMenu && IS_USER_LOGGED_IN" />
+        </div>
+      </div>
+      <div
+              id="page-wrap"
```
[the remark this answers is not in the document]
OtisRed: Yes it is. I didn't prettier it before pushing...

arturtamborski:
fair enough, its still a WIP :)

frontend/src/components/Status.vue
```diff
@@ -18,7 +18,7 @@
         <path d="M10.706 3.294A12.545 12.545 0 0 0 8 3 12.44 12.44 0 0 0 .663 5.379a.485.485 0 0 0-.048.736.518.518 0 0 0 .668.05A11.448 11.448 0 0 1 8 4c.63 0 1.249.05 1.852.148l.854-.854zM8 6c-1.905 0-3.68.56-5.166 1.526a.48.48 0 0 0-.063.745.525.525 0 0 0 .652.065 8.448 8.448 0 0 1 3.51-1.27L8 6zm2.596 1.404l.785-.785c.63.24 1.228.545 1.785.907a.482.482 0 0 1 .063.745.525.525 0 0 1-.652.065 8.462 8.462 0 0 0-1.98-.932zM8 10l.934-.933a6.454 6.454 0 0 1 2.012.637c.285.145.326.524.1.75l-.015.015a.532.532 0 0 1-.611.09A5.478 5.478 0 0 0 8 10zm4.905-4.905l.747-.747c.59.3 1.153.645 1.685 1.03a.485.485 0 0 1 .048.737.518.518 0 0 1-.668.05 11.496 11.496 0 0 0-1.812-1.07zM9.02 11.78c.238.14.236.464.04.66l-.706.706a.5.5 0 0 1-.708 0l-.707-.707c-.195-.195-.197-.518.04-.66A1.99 1.99 0 0 1 8 11.5c.373 0 .722.102 1.02.28zm4.355-9.905a.53.53 0 1 1 .75.75l-10.75 10.75a.53.53 0 0 1-.75-.75l10.75-10.75z" />
       </svg>
     </div>
-    <div v-else>
+    <div v-if="isOnline && drivesView && unsyncedDrives.length">
```
pniedzwiedzinski, line 21:
```suggestion
    <div v-else-if="drivesView && unsyncedDrives.length">
```

OtisRed: Thanks!

OtisRed:
I will close this one and pull another one from a different branch (I made some mess here and it was just easier to start fresh from a clear master fork)

Issue #413, opened 2020-01-22
Block logging for users without any group assigned


arturtamborski:
We have few admin users who shouldn't be able to log in as a driver or passenger to the regular frontend as this doesn't make sense. Let's block them from logging in by checking if said user has some is in Passengers or Drivers group.

Backend: upon login check if user has at least one group assigned and if not then raise some error (HTTP 4XX) and helpful JSON message.

Frontend: catch it, show it :)

PR #546, opened 2020-08-20
413 Block logging for users without any group assigned

Ctilhs:
[413](https://github.com/CodeForPoznan/pah-fm/issues/413)

### DIFF
--- a/backend/pah_fm/permissions.py
+++ b/backend/pah_fm/permissions.py
@@ -1,0 +1,11 @@
+from rest_framework import permissions
+from fleet_management.models import User
+
+class UserHasGroupPermission(permissions.BasePermission):
+    message = 'You are not allowed to log in because you are not associated to proper group.'
+
+    def has_permission(self, request, view):
+        try:
+            return User.objects.get(email=request.data.get('username')).groups.exists()
+        except User.DoesNotExist:
+            return False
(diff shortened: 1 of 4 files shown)


backend/pah_fm/permissions.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import permissions
+from fleet_management.models import User
+
+class UserHasGroupPermission(permissions.BasePermission):
+    message = 'You are not allowed to log in because you are not associated to proper group.'
+
+    def has_permission(self, request, view):
+        try:
+            return User.objects.get(email=request.data.get('username')).groups.exists()
```
arturtamborski, line 9:
```suggestion
            return User.objects.get(username=request.data.get('username')).groups.exists()
```
`username` is the primary key in this model.

backend/pah_fm/permissions.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import permissions
+from fleet_management.models import User
+
+class UserHasGroupPermission(permissions.BasePermission):
+    message = 'You are not allowed to log in because you are not associated to proper group.'
```
arturtamborski, line 5:
```suggestion
    message = 'You are not allowed to log in because you do not belong to any group. Please contact admins.'
```

arturtamborski (changes requested):
The permission class looks so simple and neat! I left two suggestions :)
```
would reformat /home/travis/build/CodeForPoznan/pah-fm/backend/pah_fm/permissions.py
would reformat /home/travis/build/CodeForPoznan/pah-fm/backend/pah_fm/views.py

Oh no! 💥 💔 💥
2 files would be reformatted, 54 files would be left unchanged.
```

PR #552, opened 2020-09-12
Synchronize button (fixes #422)

### LINKED ISSUE #422: 'Sync' button
I think it would be nice to add a 'Sync now' button - it's nice to be sure that it's synced

This button would dispatch the 'SYNC' action

OtisRed:
OK. I think I'm done here


frontend/src/components/SyncButton.vue
```diff
@@ -0,0 +1,96 @@
...
+    async triggerSync() {
+      this.sync();
+      this.syncStatus = 'loading';
+      this.$store.subscribe(mutation => {
+        if (mutation.type === 'drives/setData') {
+          this.syncStatus = 'success';
+          this.unsubscribe();
+        }
+      });
+    },
```
OtisRed, line 65: I have some troubles here. 

I would like to watch for mutation rather than api response. The logic responsible for commiting that mutation covers a lot of different cases and it seems unnecessary to reproduce it. So what I would want to do is:

1. trigger sync and cause "loading"
1. resolve function with "success" when mutation is committed
1. if it's not happening for 30 seconds return "fail"
1. Wait 3sec to revert button to neutral state

I tried to use setTimeout and Promise but apparently I don't know these functions well enough to make it work. If anybody has an idea how to sensibly glue it together I'll be much obliged.

OtisRed: Ok I managed to write it down. Don't bother helping with this unless you see a better solution :)

arturtamborski:
Update from Aleksandra: please leave the online-offline (red-green) dot in the corner as is.

OtisRed:
> Update from Aleksandra: please leave the online-offline (red-green) dot in the corner as is.

It's there. the button is visible only in the unlikely situation of being online and having unsynced drives. I just hope it will work as a template when it's actually needed :D

frontend/src/components/SyncButton.vue
```diff
@@ -0,0 +1,111 @@
...
+
+export default {
+  name: 'SyncButton',
+  computed: {
+    ...mapState([IS_ONLINE]),
+    ...mapGetters({ unsyncedDrives: UNSYNCHRONISED_DRIVES }),
+  },
+  data() {
+    return {
+      isLogin: this.$router.currentRoute.path === loginRoute.path,
```
arturtamborski, line 48:
```suggestion
      inLoginView: this.$router.currentRoute.path === loginRoute.path,
```
nitpick: you're using different naming conventions for the same thing.
`isLogin` and `inDrivesView` do pretty much the same thing, but are named differently, why?

double nitpick: if that convention was to be followed then the `inLoginView` should be named `inDriveListView` :P

frontend/src/components/SyncButton.vue
```diff
@@ -0,0 +1,111 @@
...
+      inDrivesView: null,
+      syncStatus: 'untouched',
```
arturtamborski, line 50: another nitpick: I guess it would be nice to use some kind of enum (or just a dict) for keeping these statuses in one place and to avoid any possible typo, but this approach works as well.

arturtamborski (approved):
Nice job, works as expected and turned out to be much smaller change than before, always a nice thing 👍🏻

frontend/src/components/SyncButton.vue
OtisRed: You are right. The `inDrivesView` condition is logically sufficient. I also refactored it to `isDrivesList` to fit `isLogin` because it was preceeding convention and I should accord to it :) TBC (to be comitted)

frontend/src/components/SyncButton.vue
OtisRed: I understand the point. I think pure numbers without meaning could be difficult to read. The dictionary idea sounds good by I don't know how this would work. I tried sth like:
`      syncStatus: {
        1: 'untouched',
        2: 'loading',
        3: 'success',
        4: 'failure',
      },`
But then v-if=syncStatus.1 was no go. It seems you cannot ascribe sheer numbers as the name of the var. 

If you got any ideas about execution I'd gladly take a suggestion. Otherwise I have no clue how to make that dict happen :P

arturtamborski: Well, to be fully correct - javascript does not support enums afaik, but you can simulate it with dict 
(example: https://stijndewitt.com/2014/01/26/enums-in-javascript/)

which turns out to your approach, but reversed:

    syncStatuses: { untouched: 1, loading: 2, success: 3, failure: 4 }

which lets you do things like:

    <div v-if="syncStatus === syncStatuses.untouched">

It's not really bulletproof, because you still can make up the states (and do typos) like: `... === syncStatuses.faiure`
but I suppose that will be less likely due to editor autocompletion of static keys.

Anyway, that's what I had in mind. The main benefit from my perspective is that by creating some king of map with these keys you're making it very obvious as to what states are actually there. If you'd use strings then it can get more difficult to track, because the possible statuses are scattered in the codebase.

OtisRed: Ok it works :)

OtisRed:
Ready to re-review :)

PR #554, opened 2020-09-16
display error message on login screen

juliaanholcer:
403 status is sent by backend when the user does not exist or when the user doesn't belong to any group. In this case, different login error is displayed. I've also changed the alert to dismissable.

### DIFF
--- a/frontend/src/store/actions.js
+++ b/frontend/src/store/actions.js
@@ -40,15 +40,19 @@ export const actions = {
 
   async [LOGIN]({ commit, dispatch }, credentials) {
     commit(mutations.SET_LOGIN_PROGRESS, true);
+    commit(mutations.SET_LOGIN_ERROR, null);
     try {
       await dispatch(`http/${HTTP_LOGIN}`, credentials);
-      commit(mutations.SET_LOGIN_ERROR, null);
       dispatch(FETCH_USER, {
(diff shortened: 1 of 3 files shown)


frontend/src/store/actions.js
```diff
@@ -40,15 +40,19 @@ export const actions = {
...
         callback: () => window.location.replace('/drive'),
       });
     } catch (err) {
       console.error(err);
-      commit(mutations.SET_LOGIN_ERROR, i18n.tc('login.login_error'));
+      const { response: {
+        status,
+      } } = err;
+
+      commit(mutations.SET_LOGIN_ERROR, status === 403 ? i18n.tc('login.login_group_error') : i18n.tc('login.login_error'));
```
pniedzwiedzinski, line 55: `403` also means wrong login/password

I found that's also problem on backend - in response details it has the warning about groups (Even if user does not exist!)

pniedzwiedzinski (changes requested):

frontend/src/store/actions.js
pniedzwiedzinski: Please check if you have the same error, we might need to create a ticket about it

arturtamborski:
I think we can just freeze the frontend (vuejs) at this point and leave it untouched until react is ready.

PR #573, opened 2021-03-03
Add transifex

arturtamborski:


frontend-react/src/App.js
```diff
@@ -5,10 +5,8 @@ import { createBrowserHistory } from 'history';
 import Container from 'react-bootstrap/Container';
 
 import routes, { renderRoutes } from './routes';
-import {
-  getMe,
-  login,
-} from './store/slices/auth';
+import { getMe, login } from './store/slices/auth';
+import LanguagePicker from "./components/LanguagePicker";
```
juliaanholcer, line 9: I need to move `eslint` config to react project, however, to avoid too many changes in the future, please use single quotes for all strings

frontend-react/src/index.js
```diff
@@ -12,6 +13,11 @@ import 'bootstrap/dist/css/bootstrap.min.css';
 
 import './index.css';
 
+tx.init({
+  token: '1/0295575608a243c1442a518c21751c8d6b1f2c53',
```
juliaanholcer, line 17: Shouldn't this be put in .env ?

frontend-react/.babelrc
```diff
@@ -1,11 +1,5 @@
...
   "presets": ["@babel/preset-react"],
-  "plugins": [
-    ["i18next-extract",
-      {
-        "locales": ["en", "uk", "so", "pl"],
-        "outputPath": "src/translations/{{locale}}/{{ns}}.json"
-      }
-    ]
-  ]
+  "plugins": []
```
juliaanholcer, line 3: As it is an empty array it canon be removed?

frontend-react/src/components/LanguagePicker.js
```diff
@@ -0,0 +1,25 @@
...
+import {useLanguages, setLocale} from '../utils/translation';
+
+const flags = {
+  ar_IQ: "IQ",
+  en: "GB",
+  ku: "IQ",
+  so: "SO",
+  pl_PL: "PL",
+  uk_UA: "UA",
+}
```
juliaanholcer, line 12: Missing semicolon

frontend-react/src/utils/translation.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+}
+
+/*** NOTE(artur):
+  Uh, ok, so the transifex live works now, kind of, but it fetches translations on every page even though 
+  it receives all translations anyway, so that's odd... there has to be a way to prefetch translations 
+  and keep them in local store some kind of middleware for this? we could load al translations for given 
+  user and see how that ends up user could have a language field for his preferred lang so that we won't 
+  have to bother with loading other langs.
+*/
```
juliaanholcer, line 39: We can figure this out - some kind of hook on init that will fetch them only when app start. We can optimize this later.

frontend-react/src/views/Home/index.js
```diff
@@ -1,16 +1,15 @@
 import React from 'react';
-import { useTranslation } from 'react-i18next';
 
 import Page from '../../components/Page';
+import useT from "../../utils/translation";
```
juliaanholcer, line 5: Please use singlequote

frontend-react/src/views/Home/index.js
```diff
@@ -1,16 +1,15 @@
...
 
 const LoginView = () => {
-  const { t } = useTranslation();
+  // I know this line sucks, but there's no better way to handle this long string now, unfortunately
```
juliaanholcer, line 7: You can try template literal with backticks

juliaanholcer (changes requested):

frontend-react/src/views/Home/index.js
arturtamborski: Nope, won't work. This stuff is rather complex, to be honest (I did investigate that to the core).

Basically `transifex` parses every `*.js` file that it finds using `babel`. It iterates trough the whole AST of every file and tries to find calls to `t()` or `useT()` (it assumes that these calls are dedicated to transifex. When it finds these calls (as in this exact line) it picks the first passed argument from the AST and checks if it's a string literal. If it is, then it collects it for pushing, but if it's anything else (even a templated literal with backticks) it fails!.

I'm working on a PR that will add support for more relaxed arguments (for example `useT("a" + "b")` does not work yet!), but I need a bit more time for that.

So until then, we're truly stuck with this approach.

frontend-react/src/utils/translation.js
arturtamborski: great!

frontend-react/src/components/LanguagePicker.js
arturtamborski: fixed!

frontend-react/src/views/Home/index.js
arturtamborski: sorry!

juliaanholcer (approved):

frontend-react/src/views/Home/index.js
arturtamborski: Not true anymore! We can replace it with concatenated const string as per this feature: https://github.com/transifex/transifex-javascript/pull/53

PR #580, opened 2021-04-09
Sidebar component

arturtamborski:
image

:D


frontend-react/src/components/LanguagePicker.js
```diff
@@ -6,6 +6,12 @@ import {
   setLocale,
 } from '../utils/translation';
 import { setLocale as setLocaleAction } from '../store/slices/ui';
+import {
+  List,
+  ListItem, ListItemIcon,
```
juliaanholcer, line 11: break this into two lines

frontend-react/src/App.js
```diff
@@ -11,19 +12,23 @@ import { createBrowserHistory } from 'history';
 import rtl from 'jss-rtl';
 import { create } from 'jss';
 import CssBaseline from '@material-ui/core/CssBaseline';
+import { Grid, IconButton, Link } from "@material-ui/core";
```
juliaanholcer, line 15: I believe we've agree to use single quotes in imports. Please change also in other places. I promise to work on and add eslint rules for this project asap :-)

frontend-react/src/components/Sidebar.js
```diff
@@ -0,0 +1,89 @@
...
+  },
+}));
+
+const Sidebar = ({ open, onClose }) => {
+  const classes = useStyles();
+  const translations = {
+    // keys are taken from routes.js::routes[].key
+    Home: useT('Home page'),
+    Login: useT('Log in'),
+    Logout: useT('Log out'),
```
juliaanholcer, line 48: To avoid capital letters  at the beginning I suggest to change keys in routes to lowercase.

Because this keys are used in two places now and there is a risk they can be changed, it would be nice to define a const to store possible values:
```
const ROUTES_KEYS = {
  HOME: 'home',
  ...
};
```
and use it both here and in routes, e.g. `[ROUTER_KEYS.HOME]: useT('Home page)`

juliaanholcer (changes requested):

frontend-react/src/App.js
arturtamborski: Oh, sorry! I often use editor autocomplete when I'm writing imports but the editor always suggests the code with double quotes and I forget to fix that.

frontend-react/src/components/Sidebar.js
arturtamborski: Thanks, will do that.

Also, small note - I tried to circumvent that problem by using `t()` function in `routes`, but it doesn't work for react... So the best bet is to define translations with `useT` hook, and that's what I did here.

juliaanholcer (approved):

Issue #583, opened 2021-04-28
Add New Drive View


vlakyi:
https://github.com/CodeForPoznan/pah-fm/blob/master/frontend/src/views/DriveFormView.vue
Recreate Add New Drive View (mostly layout, will be wired with backend later on).

juliaanholcer:
In second iteration form styling was cleared to be consistent with Login form and default Material UI look and feel. Improved validation with `yup`.

PR #584, opened 2021-04-28
Add New Drive View

vlakyi:
Add New Drive View. [(#583)](https://github.com/CodeForPoznan/pah-fm/issues/583)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Convert to LF line endings on checkout.
+*.sh text eol=lf
```
arturtamborski, line 2: :+1:

frontend-react/src/routes.js
```diff
@@ -70,6 +71,11 @@ const routes = [
     key: routeKeys.NOTFOUND,
     component: lazy(() => import('./views/NotFound'))
   },
+  {
+    path: '/drive',
+    key: routeKeys.DRIVE,
+    component: lazy(() => import('./views/Add Drive'))
```
arturtamborski, line 77: space? I don't think that's a good idea

```suggestion
    component: lazy(() => import('./views/AddDrive'))
```

frontend-react/src/views/Add Drive/index.js
```diff
@@ -0,0 +1,208 @@
+import React, { useState, useEffect } from 'react';
```
arturtamborski, line 1: as above, please rename it to `AddDrive`

frontend-react/src/views/Add Drive/index.js
```diff
@@ -0,0 +1,208 @@
...
+    car: useT('Choose a car'),
+    passenger: useT('Passenger'),
+    description: useT('Description'),
+    endLocation: useT('End location'),
+    mileageEnd: useT('Ending mileage'),
+  };
+
+  const validationSchema = yup.object().shape({
+    startLocation: yup.string().required(useT('Start location is required')),
+    mileageStart: yup.string().required(useT('Starting mileage is required')),
```
arturtamborski, line 45: why string? this doesn't align with `initialValues`

frontend-react/src/views/Add Drive/index.js
```diff
@@ -0,0 +1,208 @@
...
+                      onChange={formik.handleChange}
+                      error={!!errorText}
+                    />
+                  </WhiteBox>
+                );
+              }
+            })}
+          </Box>
+          <p>
+            {traveled} km {useT('traveled')}
```
arturtamborski, line 192:
```suggestion
            {traveled} km {traveledT}
```
please keep the `useT` out of returned JSX template, it's easier to find them when they're at the beginning of the function rather than somewhere in the returned template.

arturtamborski (changes requested):
Looks good but I have few comments.

vlakyi:
Fixed issues mentioned above in the latest commit :)

frontend-react/src/routes.js
```diff
@@ -70,6 +71,11 @@ const routes = [
     key: routeKeys.NOTFOUND,
     component: lazy(() => import('./views/NotFound'))
   },
+  {
+    path: '/drive',
+    key: routeKeys.DRIVE,
+    component: lazy(() => import('./views/AddDrive'))
```
arturtamborski, line 77: Sorry for pointing it out only now, but there's one inconsistency - this view is mapped under /drive, key is named DRIVE and the class is named DriveView, but the directory is named AddDrive. Can you rename it again to Drive?

I'm sorry for pointing it out this late but I just noticed that.

frontend-react/src/views/AddDrive/index.js
```diff
@@ -0,0 +1,209 @@
...
+                );
+              }
+            })}
+          </Box>
+          <p>
+            {traveled} km {traveledT}
+          </p>
+          <ButtonsContainer>
+            <Button type="submit" variant="contained" color="primary">
+              Submit
```
arturtamborski, line 197: these could also be marked for translation (useT)

frontend-react/src/views/AddDrive/index.js
```diff
@@ -0,0 +1,209 @@
...
+            </Button>
+            <Button variant="contained" onClick={formik.resetForm}>
+              Reset
```
arturtamborski, line 200: and here

arturtamborski (changes requested):

frontend-react/src/routes.js
vlakyi: No problem, I also was thinking about this inconsistency. Thank you for pointing this.

arturtamborski (approved):

PR #594, opened 2022-07-29
Upgrade backend to Django 4

merito:
Upgrade docker images, Django, Python dependencies.


docker-compose.yml
```diff
@@ -33,7 +35,7 @@ services:
 
   db:
     networks: [pah]
-    image: postgres:9.5
+    image: postgres:14.4
```
arturtamborski, line 38: just a note - this will cause the code to be misaligned with real DB used on prod & dev environments, so it's potentially breaking change. _Likely_ nothing should change, because we're using only basic capabilities of PostgreSQL and also ORM should cover any differences, but still, it's an important one to note.

backend/pah_fm/urls.py
```diff
@@ -33,12 +37,16 @@
...
+                                                 extra_context={'schema_url':'openapi-schema'}),
+         name='swagger-ui'),
     path("api/users/me", CurrentUserRetrieveView.as_view(), name="me"),
     path("api/passengers", PassengerListView.as_view(), name="passengers"),
     path("api/cars", CarListView.as_view(), name="cars"),
     path("api/drives", DriveView.as_view(), name="drives"),
     path("api/projects", ProjectView.as_view(), name="projects"),
     path("api/refuels", RefuelView.as_view(), name="refuels"),
+    path('api/api-token-auth/', TokenObtainPairView.as_view(), name='jwt'),
+    path('api/token/refresh/', TokenRefreshView.as_view(), name='token_refresh'),
```
arturtamborski, line 51: that's a new one?

backend/pah_fm/settings.py
```diff
@@ -21,6 +21,7 @@
 
 CSRF_COOKIE_SECURE = True
 CSRF_COOKIE_DOMAIN = BASE_URL
+CSRF_TRUSTED_ORIGINS = ["http://localhost:8080"]
```
arturtamborski, line 24: I'm wondering if this should also have 127.0.0.1 thing in it as well. Or maybe we should just ditch IP address and use localhost everywhere.

arturtamborski (approved):

PR #634, opened 2023-03-21
fix some errors with project setup

arturtamborski:


backend/Dockerfile
```diff
@@ -1,14 +1,22 @@
-FROM python:3.8-slim
+FROM python:3.10-slim
```
w1stler, line 1: Why not 3.11?

w1stler (approved):

backend/Dockerfile
arturtamborski: oh, no reason at all :P we can turn it up to 11. I'll change that

PR #652, opened 2023-07-05
Adding drives view

Cezary-Janicki:
This is my first PR, the view was done at @juliaanholcer request, and according the original view written in Vue

Drives view is a basic component that heavily utilizes material-UI v4 for easier development and maintenance. Components used are: Accordion ( for collapsible drive descriptions) and Pagination ( for pagination button on the bottom of the page).  Style-wise I was trying it keep it close to the original Drives view written in Vue. 

The full change-log is listed below.

**Changes done**
- Added Drives view 
- Changed routes in routes.js in order to accommodate for the new page being made
- Added a new route in sidebar

**Not working properly**
- User doesn't logout when logout button is clicked
- "Drives" doesn't currently have a translation

**Features to implement**
- Drives view need to fetch drives data from the back-end. Right now the view takes mock data that was downloaded from pah admin page.
- Due to implemented drives pagination, a small popup menu for number of results per page might be added


arturtamborski (approved):
you went hardcore with the mockup data! ;D
thank you for the PR.

Cezary-Janicki:
I just exported all of the data from the backend. Thanks for approving my work 😀

frontend-react/src/components/Sidebar.js
```diff
@@ -115,6 +115,7 @@ const Sidebar = ({
               className={classes.link}
               onClick={onClose}
             >
+              {console.log('traslations', r)}
```
juliaanholcer: please remove console.log

frontend-react/src/views/Drive/components/DriveForm/mockItems.js
```diff
@@ -0,0 +1,45 @@
+// values to generate MenuItems in select fields. Will be converted to redux slice.
```
juliaanholcer: This file is redundant, cars, passengers and projects are fetched from backend and real data are used in the form

frontend-react/src/views/Drives/components/DrivesList/index.js
```diff
@@ -0,0 +1,168 @@
+import React, { useState } from 'react';
```
juliaanholcer: React import is redundant as from version 17 you no longer need to explicitly import the React module in files containing React components

frontend-react/src/views/Drives/components/DrivesList/index.js
```diff
@@ -0,0 +1,168 @@
...
+    borderRadius: '10px',
+  },
+  pagination: {
+    display: 'flex',
+    justifyContent: 'center',
+    marginTop: '20px',
+  },
+}));
+
+const totalMilage = () => {
```
juliaanholcer: `totalMileage` should be calculated inside the component (using useMemo hook)
Even if we use mock data, the component should be written in such a way that in the next iteration of development, only the source of mock data will be replaced with one that will be returned by the selector from the store. So the data should be defined and operations on it should happen in the component.

frontend-react/src/views/Drives/components/DrivesList/index.js
```diff
@@ -0,0 +1,168 @@
...
+  let milage = 0;
+
+  mockDrives.map((drive) => {
```
juliaanholcer: Array.prototype.map() is a method that creates a new array populated with the results of calling a provided callback function on every element in the calling array. To achieve the effect you expect, you should use Array.prototype.reduce() method.

frontend-react/src/views/Drives/components/DrivesList/mockItems.js
```diff
@@ -0,0 +1,2804 @@
+export const mockDrives = [
```
juliaanholcer, line 1: The data format is not consistent with Drives format returned from the backend. Properties names should be in camelCase and data for nested objects (e.g. projects) should not be flattened. Mocking data in the final format allows you to avoid further changes in components when real data will be used.

frontend-react/src/views/Drives/index.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
```
juliaanholcer: react import is redundant

juliaanholcer (changes requested):
I haven't tested the solution yet, but there are some things that should be changed.

Cezary-Janicki:
Thanks for the insights. I'll make the changes when I find the time.

frontend-react/src/views/Drives/index.js
Cezary-Janicki: Removed

frontend-react/src/views/Drives/components/DrivesList/index.js
Cezary-Janicki: Thank you, removed

frontend-react/src/components/Sidebar.js
Cezary-Janicki: Sorry, removed

frontend-react/src/views/Drives/components/DrivesList/index.js
Cezary-Janicki: Rewritten to utilized reduce and use Memo

frontend-react/src/views/Drives/components/DrivesList/mockItems.js
Cezary-Janicki: This data was taken from Django Admin. Where can I check the returned back-end data so that my mock data is consistent with it?

frontend-react/src/views/Drive/components/DriveForm/mockItems.js
Cezary-Janicki: I can't recall creating this file, removed nonetheless

frontend-react/src/views/Drives/components/DrivesList/mockItems.js
juliaanholcer: I think that the best source will be to check in the browser how the data is returned by this endpoint for the vue application

